PR #25, opened 2015-01-01
Allow to turn on header-only mode without foreign code intervention

sevlat:
pugiconfig.hpp:
> // Uncomment this to switch to header-only version

I suppose it's not always convenient to edit foreign source code. This change allows to turn on header-only mode simply by defining PUGIXML_HEADER_ONLY

### DIFF
--- a/src/pugixml.hpp
+++ b/src/pugixml.hpp
@@ -19,6 +19,10 @@
 // Include user configuration file (this can define various configuration macros)
 #include "pugiconfig.hpp"
 
+#ifdef PUGIXML_HEADER_ONLY
+#include "pugixml.cpp"
+#endif
+
 #ifndef HEADER_PUGIXML_HPP
 #define HEADER_PUGIXML_HPP
 


zeux:
I tried this before; however I had to revert the change: https://github.com/zeux/pugixml/commit/cfd30d903dd77c6c46ffa2aa736968040b0eb558

The issue can likely be fixed by moving the source to pugixml.inl and including that from pugixml.cpp but I don't want to do it yet; if there are any other options I'm eager to hear them. 

Note that pugiconfig.hpp is designed to accommodate user settings - it does not have any content so it does not need to be updated when upgrading pugixml. So defining header mode in pugiconfig is not as convenient but still viable.

zeux:
This is the qmake bug report: https://bugreports.qt-project.org/browse/QTBUG-11923

zeux:
I thought of a workaround that seems to work - since qmake is not a preprocessor it can't handle macro expansion in #include directives. The change https://github.com/zeux/pugixml/commit/f3e42969a5118247de548c059e9bed69cdf208bb has new implementation of transparent header-only mode (-DPUGIXML_HEADER_ONLY is sufficient).

sevlat:
Thank you for this change. Now I use it in my project (MSVC 2013).

Issue #18, opened 2014-10-26
Double values are printed with insufficient precision


zeux:
This issue was filed on Google Code: https://code.google.com/p/pugixml/issues/detail?id=230

Ideally I would like to have perfect round-tripping of floats/doubles when using as_float/as_double. This would mean that setting double values results in much longer strings in general though since standard printf formats don't seem to support shortest-sufficiently-precise output...

At the very minimum this means extra overloads for set_value(float) to keep floats at %.9g (or whatever the correct format is).

zeux:
Fixed in 550ab4b6356ba203acb6aaebcd3d41acb9e1fa4a

mikejiang:
It would be helpful to allow the precision digits to be customized (say through `set_value` API). here is the use case: we are porting an xml file generator written in `R` into `c++` using pugixml, the default precision of casting `double` to `string` is 15, which causes the testing failed due to the mismatch between two outputs

zeux:
Yup this makes a lot of sense. Can you open a new issue for this?

PR #27, opened 2015-01-07
Add xml_text::set for float, make float/double round-trip

sdoiel61:
The issue: Double values are printed with insufficient precision #18 

Has a link to a site that offers a solution.  This patch implements that solution.

I tested the solution for floats by iterating through all values between:
std::numeric_limits<float>::min() and std::numeric_limits<float>::max(), converting to a string, then parsing the string and I obtained the value I started with.

On my machine the test for float took about 20 minutes to run on 8 cores.

I didn't attempt to run the same test with a double since I think it would take too long to run the test.

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -4608,6 +4608,13 @@ namespace pugi
 		return impl::set_value_convert(_attr->value, _attr->header, impl::xml_memory_page_value_allocated_mask, rhs);
 	}
 	
+	PUGI__FN bool xml_attribute::set_value(float rhs)
+	{
+		if (!_attr) return false;
+
+		return impl::set_value_convert(_attr->value, _attr->header, impl::xml_memory_page_value_allocated_mask, rhs);
+	}
+	
 	PUGI__FN bool xml_attribute::set_value(bool rhs)
 	{
 		if (!_attr) return false;
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -3954,10 +3954,18 @@ PUGI__NS_BEGIN
 		return set_value_buffer(dest, header, header_mask, buf);
 	}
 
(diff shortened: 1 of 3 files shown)


zeux:
Can I ask you to fix the failing test (as indicated by Travis CI on this page) and add set_value and operator= for float in xml_attribute?

sdoiel61:
I believe I have covered the cases you mentioned and fixed the large_number_test.

I'm not familiar enough with the test rig to add a test but it might be smart to add a couple of tests:

test for round trip float and double
{
        float f_orig = std::numeric_limits<float>::max();

        pugi::xml_document testDoc;
        pugi::xml_node valueNode = testDoc.append_child( "value" );
        valueNode.text().set( f_orig );
        valueNode.append_attribute( "fvalue" ).set_value( f_orig );

        float f_back_node = testDoc.child( "value" ).text().as_float();
        float f_back_attr = testDoc.child( "value" ).attribute( "fvalue" ).as_float();

       /// check to make sure f_orig == f_back_node
       /// check to make sure f_orig == f_back_attr
}

    {
        double f_orig = std::numeric_limits<double>::max();

        pugi::xml_document testDoc;
        pugi::xml_node valueNode = testDoc.append_child( "value" );
        valueNode.text().set( f_orig );
        valueNode.append_attribute( "fvalue" ).set_value( f_orig );

        double f_back_node = testDoc.child( "value" ).text().as_double();
        double f_back_attr = testDoc.child("value").attribute( "fvalue" ).as_double();

       /// check to make sure f_orig == f_back_node
       /// check to make sure f_orig == f_back_attr
    }

These simple checks only look at std::numeric_limits<float>::max() for round-trip behavior.  There might be other specific values to be checked, but these checks fail with older versions of pugixml and pass with my revisions.

zeux:
Thank you! I'll add extra tests; I just wanted to start by making sure existing builds are green.

PR #41, opened 2015-03-31
Align attributes each on separate line

halex2005:
Hi!
Here is PR with support for alignment each attribute on next line.
This alignment can simplify merge two files scenarios (for example, for compare two revisions of configuration files if only one of multiple attributes was changed).

Added `format_indent_attribute` mode:
```xml
<node
    attr="1"
    other="2">
</node>
<node
    attr="1"
    other="2" />
```

Added `format_indent_full` mode:
```xml
<node
    attr="1"
    other="2"
    >
    <sub />
</node>
<node
    attr="1"
    other="2"
    />
```

Would you like to accept this PR?
Thanks for great library!
Best regards, Aleksei.

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -3480,13 +3480,26 @@ PUGI__NS_BEGIN
 		}
 	}
 
-	PUGI__FN void node_output_attributes(xml_buffered_writer& writer, xml_node_struct* node, unsigned int flags)
+	PUGI__FN void node_output_attributes(xml_buffered_writer &writer, xml_node_struct *node, const char_t *indent, size_t indent_length, unsigned int flags, unsigned int depth)
 	{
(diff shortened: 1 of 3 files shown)


zeux:
Thinking about this further, I think it's sufficient to just implement format_indent_attribute (and always keep last attribute on the same line as the closing bracket).

This keeps the change small; it's probably good enough for making XML files more merge friendly.
Also, .NET XML has the same feature (XmlWriterSettings.NewLineOnAttributes).

Would you mind removing the extra code/flags to just keep format_indent_attribute?
You should also use text_output_indent to print indentation instead of a loop.
Also, node_output_attributes is also used for printing XML declaration; it's probably ideal to suppress this for XML declaration by or-ing the flags that are passed with format_raw in that one call.

Let me know if you want me to do these changes instead.

halex2005:
I agree with you.
Here is changes.

PR #42, opened 2015-04-22
Compact mode

zeux:
This introduces a new storage mode that dramatically reduces node size at some performance cost.

Work in progress - needs a few more fixes/tests and some micro-optimizations in various places.


zeux:
Fixes and optimizations are all in; however, I have an idea that would improve string packing for attribute (making it 8 bytes) and restore name/value pair in node_struct (which restores the size of xml_node_struct in non-compact mode to previous size but allows us to not have node_pi or has_name/value).

Since it significantly changes the diffs in non-compact portion of the code I'll have to try that before merging this or merging the new version.

PR #54, opened 2015-09-16
Style Fixes

QuickJack:
(1) A few classes should be made non-copyable
(2) C++11 semantics should be preferred to mark classes non-copyable
(3) Added explicit to several contructors in oder to prevent implicit conversions

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -237,13 +237,8 @@ PUGI__NS_BEGIN
 	template <typename T, typename D = void(*)(T*)> struct auto_deleter
 	{
 		// Non-copyable semantics
-#if __cplusplus >= 201103
-		auto_deleter(const auto_deleter&) = delete;
-		auto_deleter& operator=(const auto_deleter) = delete;
-#else
 		auto_deleter(const auto_deleter&);
 		auto_deleter& operator=(const auto_deleter);
-#endif
 
 		T* data;
 		D deleter;
@@ -2908,13 +2903,8 @@ PUGI__NS_BEGIN
 	struct xml_parser
 	{
 		// Non-copyable semantics
-#if __cplusplus >= 201103
-		xml_parser(const xml_parser&) = delete;
-		xml_parser& operator=(const xml_parser) = delete;
-#else
 		xml_parser(const xml_parser&);
 		xml_parser& operator=(const xml_parser);
-#endif
 
 		xml_allocator alloc;
 		xml_allocator* alloc_state;
@@ -3160,6 +3150,7 @@ PUGI__NS_BEGIN
 		{
 			// load into registers
 			xml_node_struct* cursor = ref_cursor;
+			char_t ch = 0;
 
 			// parse node contents, starting with question mark
 			++s;
@@ -3177,8 +3168,6 @@ PUGI__NS_BEGIN
 
 			if (declaration ? PUGI__OPTSET(parse_declaration) : PUGI__OPTSET(parse_pi))
 			{
-				char_t ch = 0;
-
 				if (declaration)
 				{
 					// disallow non top-level declarations
@@ -7450,13 +7439,8 @@ PUGI__NS_BEGIN
 	struct xpath_allocator_capture
 	{
 		// Non-copyable semantics
-#if __cplusplus >= 201103
-		xpath_allocator_capture(const xpath_allocator_capture&) = delete;
-		xpath_allocator_capture& operator=(const xpath_allocator_capture) = delete;
-#else
 		xpath_allocator_capture(const xpath_allocator_capture&);
 		xpath_allocator_capture& operator=(const xpath_allocator_capture);
-#endif
 
 		explicit xpath_allocator_capture(xpath_allocator* alloc): _target(alloc), _state(*alloc)
 		{
@@ -8380,13 +8364,8 @@ PUGI__NS_BEGIN
 	struct xpath_variable_string: xpath_variable
 	{
 		// Non-copyable semantics
-#if __cplusplus >= 201103
-		xpath_variable_string(const xpath_variable_string&) = delete;
-		xpath_variable_string& operator=(const xpath_variable_string) = delete;
-#else
 		xpath_variable_string(const xpath_variable_string&);
 		xpath_variable_string& operator=(const xpath_variable_string);
-#endif
 
 		xpath_variable_string(): xpath_variable(xpath_type_string), value(0)
 		{
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -236,6 +236,15 @@ PUGI__NS_END
 PUGI__NS_BEGIN
 	template <typename T, typename D = void(*)(T*)> struct auto_deleter
 	{
+		// Non-copyable semantics
+#if __cplusplus >= 201103
+		auto_deleter(const auto_deleter&) = delete;
+		auto_deleter& operator=(const auto_deleter) = delete;
+#else
+		auto_deleter(const auto_deleter&);
+		auto_deleter& operator=(const auto_deleter);
+#endif
+
 		T* data;
 		D deleter;
 
@@ -483,7 +492,7 @@ PUGI__NS_BEGIN
 
 	struct xml_allocator
 	{
-		xml_allocator(xml_memory_page* root): _root(root), _busy_size(root->busy_size)
+		explicit xml_allocator(xml_memory_page* root): _root(root), _busy_size(root->busy_size)
 		{
 		#ifdef PUGIXML_COMPACT
 			_hash = 0;
@@ -1030,7 +1039,7 @@ namespace pugi
 {
 	struct xml_attribute_struct
 	{
-		xml_attribute_struct(impl::xml_memory_page* page): header(page, 0), namevalue_base(0)
+		explicit xml_attribute_struct(impl::xml_memory_page* page): header(page, 0), namevalue_base(0)
 		{
 			PUGI__STATIC_ASSERT(sizeof(xml_attribute_struct) == 8);
 		}
@@ -1075,7 +1084,7 @@ namespace pugi
 {
 	struct xml_attribute_struct
 	{
-		xml_attribute_struct(impl::xml_memory_page* page): header(reinterpret_cast<uintptr_t>(page)), name(0), value(0), prev_attribute_c(0), next_attribute(0)
+		explicit xml_attribute_struct(impl::xml_memory_page* page): header(reinterpret_cast<uintptr_t>(page)), name(0), value(0), prev_attribute_c(0), next_attribute(0)
 		{
 		}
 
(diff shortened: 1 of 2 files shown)


zeux:
I don't understand the purpose of most of these changes, frankly.

Two main questions:
> (2) C++11 semantics should be preferred to mark classes non-copyable
What's the purpose of this? If you have a C++11 codebase I won't argue that C++ semantics are cleaner. pugixml is not a C++11 library though, it's just C++11 compatible.

> (3) Added explicit to several contructors in oder to prevent implicit conversions
What's the value of adding explicit to multi-argument constructors? I understand the semantics, but don't understand the value.

QuickJack:
Hi Arseny,

it is said that pugixml has some kind of support for C++11 in http://pugixml.org/news.html. The follwing define is used at several places throughout the code "#if __cplusplus >= 201103". This indicates that there are code snippets which will only get activated when using a compiler that supports at least C++11.

(2) I suggest you to read this article http://ariya.ofilabs.com/2015/01/c-class-and-preventing-object-copy.html which explains all common methods to prevent an object from beeing copyable. The big advantage of using the =delete syntax is that you will get a clear error message in case you are trying to copy a non copyable object. This would significantly simplify the usability of pugixml.
(3) The keyword explicit is ignored for contructors with multiple arguments when using a compiler that only support C++03 or earlier. C++11 and up allows to implicitly initialize such an object using {} semantics (http://stackoverflow.com/questions/13886507/explicit-on-n-ary-constructors). This could lead to difficult to track down problems.


Best regards,
Martin


Date: Thu, 17 Sep 2015 07:14:17 -0700

zeux:
> there are code snippets which will only get activated when using a compiler that supports at least C++11

Right, but they are used for extra functionality.

> The big advantage of using the =delete syntax is that you will get a clear error message in case you are trying to copy a non copyable object. This would significantly simplify the usability of pugixml.

    test.cpp:8:7: error: 'operator=' is a private member of 'pugi::xml_document'

vs

    test.cpp:8:7: error: overload resolution selected deleted operator '='

I don't see how this helps at all.

zeux:
> This could lead to difficult to track down problems.

I said I know the semantics but don't understand the value. What are the specific examples when this would lead to difficult to track down problems in cases that you change?

Also note that a lot of the changes - both deleted operators and explicit - are in internal source code. So they don't affect the ease of use or misuse of pugixml at all, they just clutter the source code in case of deleted operators.

zeux:
Also, xml_writer classes don't have to be non-copyable. I don't know exactly why you would copy them, but there's no harm in doing so.

(contrast this with all other non-copyable objects in pugixml that are non-copyable for a reason!)

QuickJack:
I did modify the xml_writer class at first. However it comply with a test rule. It now supports explicit & implicit conversions again. Please use my latest modifications from github.

Date: Fri, 18 Sep 2015 07:58:00 -0700

QuickJack:
Take for example class xml_writer_file which is copyable in your original code. The member xml_writer_file::file
 is of type void*. Copying  a class which contains pointers to some memory 
location or file handle using a default copy contructor will lead to desaster. That is because both the original object as well as the copied one would point to the same memory location. To fix this potential problem xml_writer_file should be made non copyable. Alternatively, one could write it's own copy constructor for example.

Of course, if a programmer knows exactly how to use a class like xml_writer_file, there wouldn't be a problem. However, a good code quality should guarantee that there are no such pitfalls. Applying the rules that I am talking about wouldn't clutter anything but improve both internal and external usage of pugixml.

Did you read the articles that I sent you in my first mail? Again I suggest you to read http://ariya.ofilabs.com/2015/01/c-class-and-preventing-object-copy.html. They are giving a simple example of exactly the same problem.


Date: Fri, 18 Sep 2015 07:54:33 -0700

zeux:
> Did you read the articles that I sent you in my first mail? Again I suggest you to read http://ariya.ofilabs.com/2015/01/c-class-and-preventing-object-copy.html. They are giving a simple example of exactly the same problem.

They are but their example makes zero sense. It's from a world of object oriented design with completely senseless mapping from real-world objects to classes without any explanation of a problem they're trying to solve. It almost seems to imply copying a car should behave as if a car copying in a real world does - should a copy ctor of Car generate a new serial number?..

Here's a class for you: http://en.cppreference.com/w/cpp/experimental/basic_string_view

It contains a pointer. It is copyable.

Having said that, I actually don't care that much about whether xml_writer_* is copyable or not. I'm okay with making it non-copyable - I don't think it will solve any problems but I don't think it will hurt either. My main issues are with all other changes.

QuickJack:
The problem is not
about classes that describe cars or apples but about classes that
contain pointers to memory that has been obtained using new. If such
pointers are copied we have two different objects that are referring
to the same memory. This is bad practice. There are not many cases
for which something like that can be useful in any way. By the way,
class basic_string_view is in std::experimental. Yes, there are
exceptions (nearly every rule has exception). But exceptions are not
the rule. 





The following is a
complete list of changes that I did. Please tell me exactly what
changes are problematic from your point of view. I will revert all of
these changes.




A - Changes External Interface




===========================




(1) I changed the
way uncopyable classes are marked to use C++11 semantics when a C++11
compatible compiler is used. The benefit is that the compiler will
generate a clear error message.




(2) I have marked
the following classes as uncopyable:
xml_writer_file:
Member void* file refers to memory but uses default copy ctor. 

xml_writer_stream:
Contains member of class std::basic_ostream which is non copyable
(please take a moment and think about why stream classes are not
copyable!). 

xml_document: Member
char_t* _buffer refers to memory but uses default copy ctor. 





(3) I changed the
following contructors to explicit:
xml_object_range::xml_object_range(It
b, It e) 

xml_writer_file::xml_writer_file(void*
file) 

xml_writer_stream::xml_writer_stream(std::basic_ostream<char,
std::char_traits<char> >& stream) 

xml_writer_stream::xml_writer_stream(std::basic_ostream<wchar_t,
std::char_traits<wchar_t> >& stream) 

xml_attribute_iterator::xml_attribute_iterator(xml_attribute_struct*
ref, xml_node_struct* parent) 

xml_attribute_iterator::xml_attribute_iterator(const
xml_attribute& attr, const xml_node& parent) 

xml_named_node_iterator::xml_named_node_iterator(const
xml_node& node, const char_t* name) 

xpath_variable::xpath_variable(xpath_value_type
type) 

xpath_node::xpath_node(const
xml_attribute& attribute, const xml_node& parent) 

xpath_node_set::xpath_node_set(const_iterator
begin, const_iterator end, type_t type = type_unsorted) 





B - Implementation Changes





===========================




(1) I have marked
the following classes as uncopyable
auto_deleter: Member
T* data refers to memory but uses default copy ctor. 

xml_parser: Member
xml_allocator* alloc_state refers to memory but uses default copy
ctor. 

xpath_allocator_capture:
Member xpath_allocator* _target refers to memory but uses default
copy ctor. 

xpath_variable_string:
Member char_t* value refers to memory but uses default copy ctor. 





(2) I changed the
following contructors to explicit:
xml_allocator::xml_allocator(xml_memory_page*
root) 

xml_attribute_struct::xml_attribute_struct(impl::xml_memory_page*
page)
xml_document_struct::xml_document_struct(xml_memory_page*
page) 

xml_parser::xml_parser(xml_allocator*
alloc_) 

xpath_allocator_capture::xpath_allocator_capture(xpath_allocator*
alloc) 

namespace_uri_predicate::namespace_uri_predicate(const
char_t* name) 







Date: Sat, 19 Sep 2015 09:35:07 -0700

QuickJack:
Here https://www.securecoding.cert.org/confluence/pages/viewpage.action?pageId=637 I have found another source of information that I suggest you for further reading. These guys are developing C++ coding standards for writing secure software. Under Recommendations in Rec. 09 you can find the two points that we have been discussing.

https://www.securecoding.cert.org/confluence/display/cplusplus/OOP06-CPP.+Create+a+private+copy+constructor+and+assignment+operator+for+non+copyable+objects
https://www.securecoding.cert.org/confluence/pages/viewpage.action?pageId=3416

It is also interesting to take a look at the References section for each of the articles. This should make clear that such recommendations do really make sense. 


Date: Sat, 19 Sep 2015 09:35:07 -0700

zeux:
I don't see how any of this is relevant. Yes, it's correct that when a class owns the data it should either prohibit copy construction or implement a copy constructor that performs a deep copy (or a shallow ref-counted copy). When a class does not own the data the story is different. In case of xml_writer objects, the pointers are not even heap allocated (especially for stream)!

I'll comment on the diff re: parts that need to be reverted.

src/pugixml.cpp
```diff
@@ -236,6 +236,15 @@ PUGI__NS_END
 PUGI__NS_BEGIN
 	template <typename T, typename D = void(*)(T*)> struct auto_deleter
 	{
+		// Non-copyable semantics
+#if __cplusplus >= 201103
```
zeux: As discussed in the PR, C++11 deleted operators don't really give better error messages so they don't help - please remove this #if (as well as other #ifs of the same type below).

src/pugixml.cpp
```diff
@@ -3160,6 +3177,8 @@ PUGI__NS_BEGIN
 
 			if (declaration ? PUGI__OPTSET(parse_declaration) : PUGI__OPTSET(parse_pi))
 			{
+				char_t ch = 0;
```
zeux: ch variable is special in that it's used by ENDSEG macro. Right now the closest invocation of this macro is inside this scope but the intent is to have ch variable in every function that uses the macro. So this change makes future modifications to this function slightly harder - please put this variable back.

src/pugixml.hpp
```diff
@@ -1207,34 +1238,34 @@ namespace pugi
...
 
 	public:
 		// Default constructor; constructs empty XPath node
 		xpath_node();
-		
+
 		// Construct XPath node from XML node/attribute
 		xpath_node(const xml_node& node);
-		xpath_node(const xml_attribute& attribute, const xml_node& parent);
+		explicit xpath_node(const xml_attribute& attribute, const xml_node& parent);
```
zeux: Given that xpath_node ctor with one argument is implicit, I don't see a reason for making this explicit. It can be conveniently used in C++11 occasionally e.g. for passing { attr, node }, and to keep the interface symmetric either both have to be implicit or explicit (but main ctor can't be made explicit because of compatibility - and, frankly, usability - so it does not make sense to make this one explicit).

src/pugixml.hpp
```diff
@@ -1260,22 +1291,22 @@ namespace pugi
...
 		// We define non-constant iterator to be the same as constant iterator so that various generic algorithms (i.e. boost foreach) work
 		typedef const xpath_node* iterator;
-	
+
 		// Default constructor. Constructs empty set.
 		xpath_node_set();
 
 		// Constructs a set from iterator range; data is not checked for duplicates and is not sorted according to provided type, so be careful
-		xpath_node_set(const_iterator begin, const_iterator end, type_t type = type_unsorted);
+		explicit xpath_node_set(const_iterator begin, const_iterator end, type_t type = type_unsorted);
```
zeux: A similar ctor in std::vector in C++11 is not explicit; please keep this implicit to better match the interface.

src/pugixml.cpp
QuickJack: I am using cppcheck for all of my projects. This declaration generates a warning today because it's scope can be reduced. My suggestion would be the keep this change in order to minimize the number of warnings. It would be very easy to move the declaration in the future if it makes any sense.

But anyway, I have reverted the change in order to comply with your rules.

QuickJack:
Hi Arseny,

I have reverted the changes as you requested. But now you have simply closed my pull request. What is the reason for this?


Best regards,
Martin


Date: Thu, 8 Oct 2015 21:13:05 -0700

PR #58, opened 2015-10-06
Fixed missing PUGI__FN

Breush:
Fixed a linking issue of multiple definitions in header only mode.


zeux:
Thank you! I added a test to make sure this does not happen in the future.

PR #69, opened 2015-11-30
enlarge _memory before it isn't large enough if we are working with custom malloc(different align)

guijun:
I found this issue when I'm using pugixml with my global custom allocator:
assert(reinterpret_cast(_root) + sizeof(impl::xml_document_struct) <= _memory + sizeof(_memory));

### DIFF
--- a/src/pugixml.hpp
+++ b/src/pugixml.hpp
@@ -941,7 +941,7 @@ namespace pugi
 	private:
 		char_t* _buffer;
 
-		char _memory[192];
+		char _memory[192+64];
 		
 		// Non-copyable semantics
 		xml_document(const xml_document&);


zeux:
This change would break binary compatibility on Linux; I've been hesitant to change the size of xml_document since a library linked to libpugixml1.so would break when pugixml package is updated.

Also, this change should not be required... There is a test, document_alignment, that verifies that xml_document can be allocated at any offset *as long as* the memory is pointer-aligned.

Is it possible that your allocator only guarantees 4 byte alignment, and your application is 64-bit (sizeof(void*) = 8)?

guijun:
yes, it will break binary compatibility.
yes, my application is 64-bit. I'll try to adjust alignment.

zeux:
FWIW I have a planned change that may or may not land that changes pointer encoding in nodes and attributes to be closer to that of a compact mode, which relaxes the alignment requirements for nodes/attributes to alignof(void*). This would in turn reduce the required size of _memory[] buffer.

For now, I can't increase the _memory size due to ABI concerns, and I'd rather assume pointers should be aligned to their size everywhere (although on some platforms the requirements are not as strict), so I'm closing this.

Feel free to either adjust your allocator to align returned memory to 8 bytes (generally pugixml requests blocks that are big enough so that you can add a bit of padding that may be required by your allocator to service aligned memory blocks) or change your local copy of pugixml to have a larger _memory size.

zeux:
Commit 7f91301946ce05e79fdd2ca3e79071f79fc93ba3 changes the encoding of the nodes. You should not need to change _memory any more and the allocator that returns 4 byte aligned memory on a 64-bit platform should work as long as on that platform you can perform misaligned pointer reads.

PR #79, opened 2016-01-10
Add parse_embed_pcdata flag

zeux:
This flag determines if plain character data is be stored in the parent element's value. This significantly changes the structure of the document and does not allow some documents to round-trip; this flag is only recommended for parsing documents with a lot of PCDATA nodes in a very memory-constrained environment.

Note: this PR is incomplete (missing tests and documentation)


zeux:
Note: I'm not terribly happy with this flag as it's the first flag that introduces significant differences in the DOM structure (as witnessed by the changes to various DOM functions).

It does mean that DOM can get up to 2x smaller in extreme cases (documents of the form &lt;key&gt;value&lt;/key&gt;), which may be important for embedded software or for giant documents.

annulen:
We've tested this patchset in our code base, it works fine. Thanks!

PR #82, opened 2016-01-24
Use either -std=c++03 or -std=c++11

sbeyer:
No intermediate pseudo-standards should be used when compiling.
The option -std=c++11 is used by default now.
Compile with `make defines=PUGIXML_NO_CXX11` to use -std=c++03 instead.


zeux:
-std=c++0x is used specifically for Travis as indicated by the comment in the makefile.

Your change disables testing of C++11 features on Travis.

In addition to that, my understanding is that by default gcc/clang use c++98; if there are any incompatibilities between c++03 and c++98 (not that I know of, but still), this will hide them - not sure why explicitly specifying -std is valuable?

zeux:
P.S. Is the purpose of your change to use -std=c++11 instead of c++0x?

sbeyer:
Thank you for your thoughts!
I'm going to change my branch to reflect your doubts. 

> -std=c++0x is used specifically for Travis as indicated by the comment in the makefile.

Yes, I know.

> Your change disables testing of C++11 features on Travis.

That's a good point. I did not notice that there are C++11 features in pugixml. (I only grep'ed for PUGIXML_NO_CXX11 and did not find any occurrence. However, there are occurrences of `#if __cplusplus`.) Thanks for pointing that out.

However, I think a better solution would be to keep `-std=c++11` and just use a better travis configuration.

> In addition to that, my understanding is that by default gcc/clang use c++98;
>
> if there are any incompatibilities between c++03 and c++98 (not that I know of, but still),
> this will hide them

Hm, C++98 is another good point but I thought this is not important since the standard is so old.

> not sure why explicitly specifying -std is valuable?

Because the compilers can then effectively check for standard compliance which might be important in order to work with compilers you have not tested.

For example, one commit in my branch was to add a newline, which is necessary by C++03.

> P.S. Is the purpose of your change to use -std=c++11 instead of c++0x?

The purpose is to explicitly use a standard. I'm opposed to -std=c++0x because it is a set of vaguely defined features, changing from version to version of the used compiler. Setting `-std=c++11` does also not guarantee that the compiler implementation provides all features, but it is a fully defined standard the compiler *should* comply, too.

PR #83, opened 2016-01-24
Make C++ standards testable

sbeyer:
For Makefile-based builds, C++ standards can now be set using the variable `cxxstd`.

Using this, tests for C++98, C++03, and C++11 are added to the Travis CI testset.


Makefile
```diff
@@ -12,7 +13,7 @@ EXECUTABLE=$(BUILD)/test
 VERSION=$(shell sed -n 's/.*version \(.*\).*/\1/p' src/pugiconfig.hpp)
 RELEASE=$(shell git ls-files src docs/*.html docs/*.css docs/samples docs/images scripts contrib readme.txt)
 
-CXXFLAGS=-g -Wall -Wextra -Werror -pedantic -Wundef -Wshadow -Wold-style-cast -Wcast-align
+CXXFLAGS=-g -W -Wall -Wextra -Werror -pedantic -Wundef -Wshadow -Wold-style-cast -Wcast-align
```
zeux: Is this needed? I thought -W is a legacy name for -Wextra.

Makefile
```diff
@@ -3,8 +3,9 @@ MAKEFLAGS+=-r
 
 config=debug
 defines=standard
+cxxstd=default
```
zeux: Can you default to c++11? In general if pugixml compiles in c++11 it should compile in other standards as well; AppVeyor build generally checks that it does not use C++11 feature unless they are present.

.travis.yml
```diff
@@ -10,5 +11,8 @@ env:
 script:
   - make test config=coverage defines=$DEFINES -j2
   - make test config=release defines=$DEFINES -j2
+  - make test cxxstd=c++98 -j2
+  - make test cxxstd=c++03 -j2
+  - make test cxxstd=c++11 -j2
```
zeux: This is not a good way to do this. This redundantly runs the same builds with default defines, and I don't really think it's important to test all of these on Travis. Let's remove these for now? (with defaulting to c++11 below this should retain the current testing coverage)

Makefile
sbeyer: Wow, always learning something new. That's right. I'll remove it.

.travis.yml
sbeyer: The redundancy is intended. It is useful in case someone adds C++11isms or C++03isms but compatibility to a former C++ standard should be retained.

Makefile
sbeyer: I thought `default` is the sanest option to make pugixml work anywhere (without any `-std=` option). If AppVeyor should build a C++11 build, can't we just say `cxxstd=c++11` somewhere in Appveyor.yml?

sbeyer: Just checked the AppVeyor files in the repository and it seems that AppVeyor is totally independent of the Makefile and my changes do not affect that at all?

zeux: > I thought default is the sanest option to make pugixml work anywhere (without any -std= option)

So generally I'm the only person using the Makefile :) so the default options are set up for convenience of local testing - -std=c++11 being the choice since it activates parts of pugixml code and tests that are otherwise inaccessible.

Not sure why you'd use the Makefile in other cases - it's meant as a tool to work with pugixml, not to build pugixml, since to build pugixml you could just add the .cpp to your build.

.travis.yml
zeux: I think it's a good idea to have a test for c++98; the other two are probably redundant since c++11 can be enabled by default (you can spell it out directly in the first two steps) and c++03 is a weird thing to test for in general. That'd limit the build time degradation to just having 1 extra step.

Makefile
sbeyer: > So generally I'm the only person using the Makefile :)

Ok, I thought of people who download pugixml, compile it and run the tests to see that it works. If compiling fails (and may it be only because the compiler does not know `-std=c++11`), it may distract some people of using pugixml for their projects.

That was the last point I make in favor of a "do not use `-std`"-default.
If you really want the default to be c++11, I change my branch. However, I'd say that there should still be a magic setting `cxxstd=any` or something to disable `-std` usage.

zeux: > it may distract some people of using pugixml for their projects

I've never heard of anything like that. Having an option to default to no -std usage feels fine.

zeux:
Thank you!

Issue #84, opened 2016-02-01
Add cocoapods package


igagis:
Make a cocoapods package for using with xcode and iOS

igagis:
Here is the PR:
https://github.com/zeux/pugixml/pull/85

PR #85, opened 2016-02-03
add cocoapods package

igagis:
This is the PR to fix https://github.com/zeux/pugixml/issues/84

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,22 @@
+The MIT License (MIT)
+
+Copyright (c) 2006-2015 Arseny Kapoulkine
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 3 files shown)


zeux:
My understanding is that LICENSE file is not used - is this accurate?

It's a bit of a shame that the license exists in a lot of copies, but since it already does and you have to embed it in the spec file as well, might as well not have a LICENSE file...

igagis:
When pushing podspec to cocoapods repo it validates the podspec and sources, and during that it checks if there is a license, If there  is a LICENSE file it is ok. Other approach is to embed the license right into the podspec, then it will also pass the validation even without LICENSE file.

It is preferred to have LICENSE file and not have the license embedded in podspec. The only reason  I embedded it is that for now I used 1.7 tag to package sources for cocoapods (podspec refers to 1.7 git tag) and 1.7 tag does not have the LICENSE file.

So, the embedded license should be removed from podspec when LICENSE file is merged to master and a new version tag is created.

zeux:
Ok. I don't see much of a point of having another place to put the license in at the top level since it's already in two readme files and in every source file. IMO if a package manager requires a license and allows embedding it into a configuration file that's perfect.

Thanks for the PR!

igagis:
For the same reason, I prefer to have one LICENSE file and no license in every source file or any other places. But no problem, if it is more convenient for you :).

Your're welcome!

zeux:
Yeah, I'd very much prefer just once license boilerplate but unfortunately people have asked me before to add it verbatim to sources so that when they embed the sources in their codebase it's obvious that it's legal, it has to be in documentation and I'm not sure if documentation formats I'm using support file includes, it has to be there in the README.md since people want to know, etc. etc. :)

Issue #73, opened 2015-12-17
std::string and std::string_view support in the same way for setters and constructors as for const char *


biodranik:
Hi and thanks for amazing library!

Do you plan to add std::string and std::string_view overloads for most setters/constructors? It makes code usage much more convenient.

DasNaCl:
This might become much more interesting when the standard library has std::string_view

zeux:
Here's some previous discussion on the topic: https://code.google.com/p/pugixml/issues/detail?id=49

That thread also has different implementations of a possible solution.

zeux:
I experimented a bit with string_arg approach (the link above appears to be dead, but basically one way to implement this is to replace all existing `const char_t*` arguments with `string_arg` which would be a custom wrapper type for a char_t pointer with implicit ctor from `const char_t*` and `const std::basic_string<char_t>&`.

The biggest issue apart from breaking binary compatibility (which would mean I'd like to defer this until pugixml 2.0 to avoid potential issues with existing applications that link to pugixml dynamically) is the overload issues with `xml_attribute::set_value` / `xml_text::set`, where `bool` overload is now selected instead of `string_arg` one (which is probably fixable by keeping the `const char_t*` overload...)

Overall this seems like a good solution. There's obviously extra overhead in debug builds but there is no cost if compiler inlines the helper (which it should when optimizations are enabled).

arturbac:
> Do you plan to add std::string overloads for most setters/constructors? It makes code usage much more convenient.

It would be better to support std::string_view, it will cover any strings and will solve problem with using unterminated strings.
Actualy with modern project which uses std::string_view or boost::string_view instead of raw char const * every time i need to pass data to pugi i have to create temporary string for null termination

davidhunter22:
A smaller but still useful change would be to have a 
`bool xml_node::set_value(const char_t* rhs, size_t strlength );`
and just not do the impl::strlength(rhs) call in the implementation. The existing set_value can just be
`bool xml_node::set_value(const char_t* rhs ) { return set_value( rhs, impl::strlength(rhs) ); }`

air2:
I think std::(w)string_view is much more desirable as std::(w)string. And constructing a string_view from a std::string is very cheap, however constructing a string from a string_view is very expensive, so I would skip the std::string implementation and favor the std::string_view implementation.

Sailanarmo:
[...]

PR #88, opened 2016-02-20
stl: add ability to set/get pugi::string_t to pugi elements

stanionascu:
This allows using pugi with more convenience
e.g.: if (node.sname() == "foobar");

### DIFF
--- a/src/pugixml.hpp
+++ b/src/pugixml.hpp
@@ -335,8 +335,17 @@ namespace pugi
 		const char_t* name() const;
 		const char_t* value() const;
 
+	#ifndef PUGIXML_NO_STL
+		string_t sname() const { return name(); }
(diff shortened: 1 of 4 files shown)


zeux:
This is covered in issue #73 (wrt setters).

For getters as well, I don't like the idea of having extra interfaces - the ideal solution would be to expand the solution to #73 to cover getters as well. The proposed patch focuses on name/value but there are a lot of other functions (xml_text, xml_attribute::as_string, xml_node::child_value) - making a std::string version for all of them becomes unwieldy.

For now there are suitable workarounds that involve a bit of typing - std::string(node.name()), or node.name() == "foobar"s for C++14; I'd prefer to wait until a complete solution otherwise. So I believe it's premature to merge this at this point.

PR #90, opened 2016-02-25
Put CMakeLists.txt in the project root.

jippeholwerda:
Having CMakeLists.txt in the project root makes it so much easier to use pugixml
as an external dependency in another CMake project.


snikulov:
@zeux  :+1: I also vote for this change. 
It'll help to re-use your project easily with External_ProjectAdd.

zeux:
There's a small issue with putting CMakeLists.txt in the root - running cmake will probably overwrite the default Makefile. Also I guess some Linux distributions will have to adjust their packaging scripts as a result of this change (although I could temporarily leave CMakeLists.txt in scripts/ folder, and remove it in a following version).

Regardless, this seems like a good change. I'll test this a bit before merging.

snikulov:
@zeux you can enforce out-of-source build, as clang do and as I've implemented for splunk/pion to avoid default makefile overwrite.

snikulov:
Here the [example](https://github.com/splunk/pion/blob/develop/cmake/PionUtils.cmake#L50)

rolftimmermans:
Any news on this? Thanks!

PR #92, opened 2016-03-24
Do not assume that fclose can be converted to int(*)(FILE*)

QUSpilPrgm:
Do not assume that fclose can be converted to int(* )(FILE*) because some compilers use a special calling convention for stdlib functions like fclose.

This change is necessary to make pugixml successfully compile as part of our game project where we use [vectorcall](https://msdn.microsoft.com/en-us/library/dn375768.aspx) as default calling convention on x64. The function fclose is explicitly marked as cdecl which makes the function incompatible to the function pointer type int(* )(FILE*).

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -255,6 +255,13 @@ PUGI__NS_BEGIN
 			return result;
 		}
 	};
+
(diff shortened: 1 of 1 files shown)


zeux:
Thanks, merged!

Issue #87, opened 2016-02-20
Extra space in empty node


JayXon:
`<node/>` will become `<node />` even if format_raw is set.
Is this intended?

zeux:
This is intended.

pugixml does not keep internal whitespace inside nodes for performance/memory reasons. So format_raw emits XML with a fixed markup which for I guess style/historical reasons ended up having a space before /.

JayXon:
Thanks for the response.
IMHO people would care more about XML size than style if they are using format_raw since it's usually used in serialization and it's not intended to be read by humans.

zeux:
Right, so there is no way to preserve the formatting in general - but it is of course possible to change format_raw output to always remove the space.

It does make sense, I guess - this may be literally the only character that's printed in format_raw mode that is not *required* to be there for XML to round-trip... I'll look into this a bit more.

zeux:
Changing this breaks 110 unit tests :) which is not surprising and easy to fix.
I have no way of knowing whether any users depend on the existing behavior. Most likely the answer is "no", and there have been some minor formatting changes in the past like that.

Libraries that always add a space:
* .NET (XmlWriter)
* TinyXML
* Python xml.etree.ElementTree

Libraries that don't add a space:
* libxml2
* rapidxml
* TinyXML2
* Python xml.dom.minidom

PR #95, opened 2016-04-13
Add flag format_no_empty_element_space to omit space before end of self-closing element tags

rolftimmermans:
Empty elements are serialised to `<element />`. The space is not necessary for most XML parsers. Unfortunately it's not configurable. So to get the minimal XML size, it could be useful to omit the space.

I've added the option `format_no_empty_element_space` to control this.

I'd much prefer the omission of the space to be the default, but obviously that could break existing code.


zeux:
I actually already have the change that removes the space if you use format_raw. Not too worried about this breaking code - format_raw is not used as often as other types of formatting, and I can't see an application relying on the extra space.

rolftimmermans:
Well that works for me! 😄

zeux:
See #87.

PR #108, opened 2016-08-04
Expand Nuget Configuration Options

### LINKED ISSUE #107: Nuget Package
I was wondering if there was anyway you guys could make a few more options for your Nuget packages. I am currently in need of a VS2013 build with runtime library of MT and MTd. I am not sure what the effort involved in this would be. I can attempt to learn how to build these packages and contribute, too if needed.

Thanks!

steji113:
I have attempted to expand the Nuget configuration to include VS 2013 static builds in addition to the VS 2015 that currently exists.

Closes #107

### DIFF
--- a/scripts/nuget.autopkg
+++ b/scripts/nuget.autopkg
@@ -1,3 +1,11 @@
+configurations {
+      	// This node contains custom pivot information.
+      	Toolset 
+      	{
+           	key: "PlatformToolset";
+           	choices: { v140, v120, v110, v100 };
+      	}
+}
 nuget {
 	nuspec {
 		id = pugixml;
--- a/scripts/nuget.autopkg
+++ b/scripts/nuget.autopkg
@@ -24,9 +24,13 @@ nuget {
 	files {
 		include: { "..\src\*.hpp" };
 		
-		[x86,release] { lib: vs2015\Win32_Release\pugixml.lib; }
-		[x86,debug] { lib: vs2015\Win32_Debug\pugixml.lib; }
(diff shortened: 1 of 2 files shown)


steji113:
@zeux It seems that this worked and generated the builds needed. Would you be able to upload this modified nuget package?

scripts/nuget.autopkg
```diff
@@ -1,3 +1,11 @@
+configurations {
+      	// This node contains custom pivot information.
+      	Toolset 
+      	{
+           	key: "PlatformToolset";
+           	choices: { v140, v120, v110, v100 };
```
zeux: I'm not very familiar with NuGet - this list has 4 versions of VS but below only 2 of them have associated files - does this list have to have just v140 and v120?

zeux:
Beyond the comment above I can't easily upload a package for 1.7, since master already contains many changes beyond that. I could upload a 1.8 pre-release (after changing the version number I guess), although I need to look at the full list of changes since 1.7 to figure out whether current master is close to 1.8 or not.

scripts/nuget.autopkg
steji113: @zeux I am also quite unfamiliar with NuGet. I had found this on StackOverflow as a workaround for the fact that the CoApp NuGet autopkg doesn't have the configuration included for VS 2015 yet. It probably will work either way.

steji113:
@zeux What if a branch was made based off of the 1.7 release and I made just the NuGet changes there? If AppVeyor runs the build based off of that to generate the package, would that suffice? Otherwise, do you know what the timing of the 1.8 release is looking like?

PR #109, opened 2016-08-08
Remove static pivot from NuGet autopkg.

steji113:
This pivot seemed to be causing issues because the NuGet package thought it was actually dynamic.

Refs #107

### DIFF
--- a/scripts/nuget.autopkg
+++ b/scripts/nuget.autopkg
@@ -32,13 +32,13 @@ nuget {
 	files {
 		include: { "..\src\*.hpp" };
 		
-		[x86,v120,static,release] { lib: vs2013\x32\pugixmls.lib; }
-		[x86,v120,static,debug] { lib: vs2013\x32\pugixmlsd.lib; }
-		[x64,v120,static,release] { lib: vs2013\x64\pugixmls.lib; }
-		[x64,v120,static,debug] { lib: vs2013\x64\pugixmlsd.lib; }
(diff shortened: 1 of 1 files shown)


zeux:
Hmm - this is strange.

According to http://coapp.org/reference/autopackage-ref.html, static = static library; not sure what the default is but if it's dynamic, it'd try to link pugixml as DLL which it's not?

Can you clarify on what issue you're currently seeing if you keep static? The original NuGet package used static for VS2015 and it worked fine there.

zeux:
I guess maybe it does not matter since the linker will correctly link in the static library regardless of what NuGet thinks, and since we only really have one option for this pivot it is probably reduntant.

steji113:
In the pugixml.targets file that is automatically generated it sets the linkage to dynamic, but then the conditions for linking were looking for it to be set to static. The documentation from CoApp.com is quite poor in this area :(

steji113:
BTW, I put this generated package in my solution and it seemed to fix the issue!

zeux:
I've uploaded pugixml 1.7.2 to nuget with this fix

steji113:
@zeux Thank you for all of your assistance on this. It worked beautifully on VS 2013 Win32 - Release / Debug and x64 - Release / Debug.

zeux:
Verified it works for VS2015 as well, so the Appveyor build is working well which is great. I'll keep the nuget-1.7 branch around for now in case I need to add any more VS versions/configurations. Thanks for the pull requests!

steji113:
Absolutely!

PR #111, opened 2016-08-22
Add Symbols to NuGet Package

### LINKED ISSUE #110: NuGet Package - PDB Files
The current NuGet package only supplies the static libraries but does not include .pdb files to aid in debugging. I am proposing to include the PDBs in the NuGet package and uploading it alongside the current version to assist developers troubleshoot issues.

Thanks!

steji113:
Added debug symbols to the NuGet package for developers looking to step into pugixml code.

Closes #110

### DIFF
--- a/scripts/nuget.autopkg
+++ b/scripts/nuget.autopkg
@@ -32,27 +32,35 @@ nuget {
 		
 		[x86,v120,release] {
 			lib: vs2013\x32\pugixmls.lib;
+			symbols: vs2013\x32\pugixmls.pdb;
 		}
 		[x86,v120,debug] {
 			lib: vs2013\x32\pugixmlsd.lib;
+			symbols: vs2013\x32\pugixmlsd.pdb;
 		}
 		[x64,v120,release] {
 			lib: vs2013\x64\pugixmls.lib;
+			symbols: vs2013\x64\pugixmls.pdb;
 		}
 		[x64,v120,debug] {
 			lib: vs2013\x64\pugixmlsd.lib; 
+			symbols: vs2013\x64\pugixmlsd.pdb;
 		}
 		[x86,v140,release] {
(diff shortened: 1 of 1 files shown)


zeux:
I have vague distant memories of trying to do this before and NuGet/coapp really wanting to submit a separate package and then trying to build it and it not working...

Does this change result in debugging working *after* you remove the original PDBs in the build folders? Is there just one package being generated, or two?

steji113:
So I do believe this will require two NuGet packages that would be generated, one for the libs and one for the symbols. Apparently this is the normal mode of operation as it makes the symbols optional for people who might not want them.

Also, as of right now it does not seem to be grabbing any PDB files from the AppVeyor build so I am not completely sure if this will be working correctly. I chose where the PDBs should be based off of the VS project properties. Is there anyway to see all of the output of an AppVeyor build?

zeux:
I don't think you can see the output of AppVeyor build anywhere, but you can probably tag files that you care about as artifacts in appveyor.yml.

So, for debugging you generally need two things:

- Debug information
- Source

I looked into this a bit closer and realized I already solved the "debug information" part of the problem for VS2015 NuGet packages by using old-style debug information (this is a setting you can switch in project files). I'm guessing VS2013 projects don't use that. This embeds the information in the .obj/.lib files, making them a bit bigger; the linker either discards this (if you link without debug info) or merges the debug info into the master .pdb file (if you link with debug info).

As far as the source files go, I don't remember if there was a way to put them into NuGet package (and have the debugger automatically find them) - I probably assumed that people who need to debug can grab them from GitHub.

zeux:
Closing since it seems that the correct solution here is to switch to /Z7 for VS2013; let me know if this does not work for whatever reason, and feel free to open a PR with VS2013 project file changes if necessary.

PR #115, opened 2016-09-12
Adding PKG-CONFIG support

peterbud:
Hi Zeux, 

I have added a basic support for PKG-CONFIG file, which makes life much easier on non-Windows platforms.
Also made minor modfification to create a separate install directory under INCLUDE and LIB directories, I believe those should not break MSVC builds (I'm building on Windows but under MSYS2).

LMK if this works for you.

Best regards,

Peter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 build/
+.vscode/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,11 @@ endif()
 include(GNUInstallDirs)
 mark_as_advanced(CLEAR CMAKE_INSTALL_LIBDIR CMAKE_INSTALL_INCLUDEDIR)
 
+set(INSTALL_BIN_DIR "bin" CACHE PATH "Installation directory for executables")
+set(INSTALL_LIB_DIR "lib" CACHE PATH "Installation directory for libraries")
+set(INSTALL_INCLUDE_DIR "include" CACHE PATH "Installation directory for header files")
+set(INSTALL_PKGCONFIG_DIR "lib/pkgconfig" CACHE PATH "Installation directory for pkg-config file")
+
 set(HEADERS src/pugixml.hpp src/pugiconfig.hpp)
 set(SOURCES ${HEADERS} src/pugixml.cpp)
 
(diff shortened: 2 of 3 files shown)


zeux:
One part of the change I have reservations about is the change of the default install behavior. Previously you could do cmake && make && sudo make install and then #include "pugixml.hpp" would work since it would find the header in the default include path.

Of course this came at the cost of only allowing to install one version by default, but still - I feel like this is a usability issue.

If this is required by PKGCONFIG, maybe it makes sense to have a special CMake define (e.g. BUILD_PKGCONFIG) that reconfigures the build to be pkg-config compatible, including extra build step and include/libdir changes?

peterbud:
I understand you, and makes sense what you say. In MSYS and Linux that's no problem as PKGCONFIG gives you the INCLUDE and LIB directories :) 
Let me rework this as an options in CMAKE to put include files and lib in separate directories (default OFF): in that way default compile would work as before. 
Are you OK to generating PKGCONFIG file always, or you would like to have that also driven by a CMAKE option?

vdidenko:
Keeping the `cmake && make && sudo make install` is highly preferable as an expected simple default for potential users. Simplifies containerized operations among other things.

peterbud:
Here is the updated version with two additional options. With the first you can turn on the PKGCONFIG generation, and with the second one you can install include and lib file into a separate direcotry. By default both are turned off.

zeux:
Thanks! I want to test this a bit to make sure I understand how to use it and what pkg-config is about, will try to merge this in the next few days.

zeux:
I'm not quite clear on the purpose of INSTALL_LIB_DIR et al - it seems they are only used for pkg-config. I plan on merging this and reworking to keep just BUILD_PKGCONFIG that will simultaneously change the install folder to a versioned name and generate pkgconfig files - any objections?

zeux:
I prepared pkg-config branch: https://github.com/zeux/pugixml/tree/pkg-config that has this PR + some changes on top of that intended to simplify the flow; there's just one define now, BUILD_PKGCONFIG, and also it works regardless of whether the library is shared or static.

Let me know if this works for you and I'll merge to master.

peterbud:
Well, this looks much better than my version :) Thank you, works just perfectly!

PR #117, opened 2016-09-22
Fix broken link at quickstart page

digawp:
Should be manual.html#install.building instead of manual/install.html#install.building


zeux:
Thanks!

PR #119, opened 2016-10-12
Add C++11 'override' specifier

pavelkryukov:
In a member function declaration or definition, 'override' ensures that the function is virtual and is overriding a virtual function from the base class. The program is ill-formed (a compile-time error is generated) if this is not true. Since version 5.1, GCC supports flag '-Wsuggest-override', which warns if 'override' specifier is misssing. We've enabled this flag in our project and found that the specifier is missing in pugixml headers. Although we've disabled the warning for files including pugixml, it might be useful to add 'override' speficiers for headers as we suggest in this pull request.

### DIFF
--- a/src/pugixml.hpp
+++ b/src/pugixml.hpp
@@ -72,6 +72,15 @@
 #	endif
 #endif
 
+// If no API for functions is defined, assume default
+#ifndef PUGIXML_OVERRIDE
+#	if __cplusplus >= 201103
+#		define PUGIXML_OVERRIDE override
+#	else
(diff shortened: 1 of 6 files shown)


src/pugixml.hpp
```diff
@@ -72,6 +72,15 @@
 #	endif
 #endif
 
+// If no API for functions is defined, assume default
```
zeux: This comment seems to be copied from elsewhere in the header and is incorrect

pavelkryukov: Sorry, my fault. Added a new comment instead of this.

PR #120, opened 2016-10-26
fixed gcc build ptrdiff_t should be in std namespace

evgen48:
The compilation of pugixml started to fail for me due to prdiff_t error.
.../thirdparty/pugixml/src/pugixml.hpp:771:11: error: ‘ptrdiff_t’ does not name a type
   typedef ptrdiff_t difference_type;
           ^
So I've checked that c++ version of ptrdiff_t should has std namespace, unless you are compiling c code. However pugi isn't c - friendly so we have to follow c++ doc.
http://en.cppreference.com/w/cpp/types/ptrdiff_t

Signed-off-by: Eugene Smirnov o1o2o3o4o5@gmail.com

### DIFF
--- a/src/pugixml.hpp
+++ b/src/pugixml.hpp
@@ -651,7 +651,7 @@ namespace pugi
 		xml_object_range<xml_attribute_iterator> attributes() const;
 
 		// Get node offset in parsed file/string (in char_t units) for debugging purposes
-		ptrdiff_t offset_debug() const;
+		std::ptrdiff_t offset_debug() const;
 
 		// Get hash value (unique for handles to the same object)
 		size_t hash_value() const;
@@ -768,7 +768,7 @@ namespace pugi
 
 	public:
 		// Iterator traits
-		typedef ptrdiff_t difference_type;
+		typedef std::ptrdiff_t difference_type;
 		typedef xml_node value_type;
 		typedef xml_node* pointer;
 		typedef xml_node& reference;
@@ -810,7 +810,7 @@ namespace pugi
 
 	public:
 		// Iterator traits
-		typedef ptrdiff_t difference_type;
+		typedef std::ptrdiff_t difference_type;
 		typedef xml_attribute value_type;
 		typedef xml_attribute* pointer;
 		typedef xml_attribute& reference;
@@ -846,7 +846,7 @@ namespace pugi
 
 	public:
 		// Iterator traits
-		typedef ptrdiff_t difference_type;
+		typedef std::ptrdiff_t difference_type;
 		typedef xml_node value_type;
 		typedef xml_node* pointer;
 		typedef xml_node& reference;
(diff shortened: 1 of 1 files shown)


zeux:
Can you clarify which GCC version and which platform this is? pugixml intentionally includes C header `<stddef.h>` to get a ptrdiff_t that's not in std namespace - not all compilers have std::ptrdiff_t (I think Borland may be one that fails)

evgen48:
I've narrowed down the issue to following  include sequence,
```
#include <cxxabi.h>
#include "pugixml.hpp"
```

withoud cxxabi.h no problems appeared. Since this header is gcc specific, probably there problem as well. 
gcc version 5.4.1 20160904 (Ubuntu 5.4.1-2ubuntu1~14.04)

zeux:
I can't reproduce this on gcc 5.4.0 20160609. Let me see if I can upgrade my Ubuntu version to get 5.4.1.

FWIW setting _GLIBCXX_NAMESPACE_C define before including `cxxabi.h` may resolve this, but I'd like to get to the bottom of this.

zeux:
Can't reproduce with gcc version 5.4.1 20160929 (Ubuntu 5.4.1-2ubuntu2) either.

What compilation flags do you use?

evgen48:
Thanks @zeux  for looking into this, finally I was able to find a problem in my code cxxabi.h was mistakenly included under namespace, and for some reason everything gets compiled besides pugixml,  I doubt now that this should be fixed. It is up to you. here is a reproducer. Note if to remove first line that cannot be compiled anyway.

```
#include <memory>

namespace MyNamespace {
	#include <cxxabi.h>
}

#include <pugixml.hpp>

int main() {
	return 0;
}

```
/usr/bin/c++   -g -fPIE   -fvisibility=default -std=gnu++11 -o CMakeFiles/PugiTests.dir/pugitest.cpp.o -c pugitest.cpp
In file included from pugitest.cpp:7:0:
../thirdparty/pugixml/src/pugixml.hpp:654:3: error: ‘ptrdiff_t’ does not name a type
   ptrdiff_t offset_debug() const;
   ^
../thirdparty/pugixml/src/pugixml.hpp:771:11: error: ‘ptrdiff_t’ does not name a type
   typedef ptrdiff_t difference_type;
           ^
../thirdparty/pugixml/src/pugixml.hpp:813:11: error: ‘ptrdiff_t’ does not name a type
   typedef ptrdiff_t difference_type;
           ^
../thirdparty/pugixml/src/pugixml.hpp:849:11: error: ‘ptrdiff_t’ does not name a type
   typedef ptrdiff_t difference_type;
           ^
../thirdparty/pugixml/src/pugixml.hpp:945:3: error: ‘ptrdiff_t’ does not name a type
   ptrdiff_t offset;
   ^
../thirdparty/pugixml/src/pugixml.hpp:1050:3: error: ‘ptrdiff_t’ does not name a type
   ptrdiff_t offset;

zeux:
Ah, okay - this makes sense. cxxabi.h includes stddef.h which hides the ptrdiff_t symbol inside the namespace (I'm somewhat surprised the std namespace does not suffer the same issue). I believe there's no action needed on this PR then, so I'll close it. Thanks for investigation!

PR #121, opened 2016-11-05
Add VS2013 check for C++11 availability

ifarbod:
VS 2013 supports C++11, but __cplusplus macro hasn't updated, and its value is still 199711 (C++98?) so the old check always fails, even though the compiler supports C++11. Fun fact is they haven't even updated it in the most recent VS versions that i'm currently using (2015 w/ update 3 and 15 w/ Preview 5)

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -11923,7 +11923,7 @@ namespace pugi
 		}
 	}
 
-#if __cplusplus >= 201103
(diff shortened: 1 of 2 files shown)


ifarbod:
Looks like the travis ci build jobs failed.

zeux:
Thanks! Merged, and relaxed the MSVC version requirement to VS2010 since move semantics support is roughly equivalent throughout 2010-2013.

ifarbod:
Great!

PR #130, opened 2017-01-17
Added target_include_directories() to export include directories

### LINKED ISSUE #126: imported cmake target does not contain include directory
Instead of just linking pugixml in a using CMakeLists.txt File by
`target_link_libraries(myExe pugixml)`
I addtionally have to set the include path of pugixml
`target_include_directories(myExe ${pathtopugiinclude})`, which (pathtopugiinclude) I have to manually set (it can be derived from Imported_location, but that is tedious)

If the following line:
`target_include_directories(pugixml $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}${INSTALL_SUFFIX}>)`
was added in the CMakeLists.txt around line 58 before 
`install(EXPORT pugixml-config DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/pugixml)`
and after 
`if(BUILD_PKGCONFIG)
	# Install library into its own directory under LIBDIR
	set(INSTALL_SUFFIX /pugixml-${PUGIXML_VERSION_STRING})
endif()`

Now `target_link_libraries(myExe pugixml)` would be enough....

tested with pugixml 1.7 and 1.8 relases

dknibbe:
This closes #126


zeux:
Thanks!

Orphis:
This doesn't work with CMake 2.6. `target_include_directories` was added in 2.8.12.
I suggest raising the minimum version to something modern past version 3 to get rid of some cruft in your build scripts.

zeux:
Thanks - I updated the minimum CMake version to 2.8.12.

PR #134, opened 2017-03-05
Silence g++ 7.0.1 -Wimplicit-fallthrough warnings

sbeyer:
This is accomplished by explicitly marking the fallthrough
using the C++17's `[[fallthrough]]` attribute.
Although `[[fallthrough]]` is C++17, it is supported by
clang++ and g++ >= 7 even if you use, e.g., -std=c++11.
This is a good thing because it allows us to easily silence
these warnings regardless of the used -std flag. (Otherwise
we would have to use `__attribute__((fallthrough))` for g++;
luckily there is no need to do so.)

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -10110,10 +10110,9 @@ PUGI__NS_BEGIN
 
 				if (_rettype == xpath_type_boolean)
 					return _data.variable->get_boolean();
-
-				// fallthrough to type conversion
 			}
 
+			// fallthrough
 			default:
 			{
 				switch (_rettype)
@@ -10246,10 +10245,9 @@ PUGI__NS_BEGIN
 
 				if (_rettype == xpath_type_number)
 					return _data.variable->get_number();
-
-				// fallthrough to type conversion
 			}
 
+			// fallthrough
 			default:
 			{
 				switch (_rettype)
@@ -10538,10 +10536,9 @@ PUGI__NS_BEGIN
 
 				if (_rettype == xpath_type_string)
 					return xpath_string::from_const(_data.variable->get_string());
-
-				// fallthrough to type conversion
 			}
 
+			// fallthrough
 			default:
 			{
 				switch (_rettype)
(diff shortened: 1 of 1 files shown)


zeux:
My understanding is that this change would break compiling with -ansi: https://godbolt.org/g/wwm9Rs

A few alternative options that I see to fix this:

- Move ast_variable case inside default (just use an if, e.g. `if (_type == ast_variable && _rettype == xpath_type_boolean) ...;`)
- Extract type conversion logic into a separate member function, e.g. `eval_boolean_convert` and the like, and call it from both ast_variable and default cases)

jay:
I am not sure if this is related but in curl which is C there are a few `/* FALLTHROUGH */` to signal intended fallthroughs and apparently it works to silence gcc7, according to a recent PR. See curl/curl@cbff751

zeux:
@jay Interesting; this does seem to work in gcc7 - appears to work for both C-style and C++-style comments, as long as fallthrough is the entirety of the comment text (but it can be lowercase). So I guess a simpler short-term fix for the warning would be to change the existing "fallthrough to type conversion" comment to just say "fallthrough".

sbeyer:
Yes, the other variants are possible and are probably better than adding the keyword.

However, I checked with several g++ and clang++ versions and different -std flags, as well as with MSVC which simply ignores the setting. I don't know what godbolt is doing; my g++ 7.0.1 is only some days older.
I am not sure why `__has_cpp_attribute(fallthrough)` is true when `[[fallthrough]]` is not available?

The `/* FALLTHROUGH */` switch does not work. Also, the lines were already commented by some `// fallthrough [...]` comment which should also be acknowledged by the compiler if `/* FALLTHROUGH */` worked.

zeux:
godbolt fallthrough example in C++: https://godbolt.org/g/wAekCz

sbeyer:
Okay it seems the fallthrough code in g++7 is under heavy development ;D

zeux:
Ah, but it only works if it's in the same scope as the case statement, so the comment would have to stay outside of braces.

sbeyer:
You don't need the scope anyway in many cases. Do you have them as some kind of coding style?

sbeyer:
I just updated my PR to use the simplest change. The information "to type conversion" is lost.

zeux:
Thank you!

I think braces are here for symmetry with other complicated (multi-line) cases, most of which require variable declaration which require scopes. There's no specific code style rule for pugixml that defines whether they should be used in this case.

PR #145, opened 2017-06-04
use snprintf instead of sprintf

noresources:


jay:
Note that many versions of the Microsoft CRT don't have snprintf, though they do have _snprintf which is slightly different because it doesn't append a null terminator when the formatted output is >= len.

zeux:
At least one other compiler toolchain in the extended test suite I'm using for compatibility checking does not support snprintf (Digital Mars C++), so I'm slightly concerned with portability.

What is the rationale behind this change?

noresources:
We are banning any use of unsafe functions in our code base. I admit that in the case of pugixml, the buffers *should be* always tall enough to hold the result. It's just a good practice to avoid sprintf.
For Digital Mars, It seems it use _snprintf but I have no way to test it myself.

zeux:
@noresources Is your codebase compiled in C++11 mode (-std=c++11)? If so one option would be to use snprintf if it's known to be available (which is the case when __cplusplus >= 201103 or when _MSC_VER >= 1900) - in which case we also wouldn't need to manually zero-terminate the result.

zeux:
Thanks!

Issue #146, opened 2017-06-12
Add VS2017 project files and v141 NuGet package


igagis:
Add VS2017 project files and include binaries built with tools v141 to nuget package.

igagis:
Pull request https://github.com/zeux/pugixml/pull/147

PR #147, opened 2017-06-15
VS2017 project

igagis:
Fix for https://github.com/zeux/pugixml/issues/146


zeux:
@igagis My understanding is that you had to convert the nuget_build.bat to PS which is fine - but what I don't understand is why we need two images now? Is it possible to just use the VS2017 image?

igagis:
This is becaus VS2017 image has only VS2015 and VS2017 installed, while VS2015 image has all VS'es except VS2017. So, building on VS2017 image is ok, since we only build with v140 and v141 tools and pack results to nuget, but testing is done on all previous VS versions, so we do building on VS2017 image and testing on VS2015 image.

PR #149, opened 2017-06-16
Improve code coverage

zeux:
This change expands coverage by excluding assert(false) statements from codecov report and by adding tests for loading files with unusual behavior to trigger ftell/fread issues.


tuananh:
codecov/patch 100% of diff hit 😆

PR #183, opened 2018-02-27
Add noexcept specifiers to move special members where possible

brandl-muc:
Having move special members is only half the deal. If they are not noexcept then for example std::vector::push_back will copy instead of move to keep giving the strong exception guarantee.
Therefore this PR adds noexcept to those special move members that can be made noexcept.

Also see these coding guidelines:
- [Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-move-noexcept)
- [High Integrity C++ Coding Standard](http://www.codingstandard.com/rule/12-5-4-declare-noexcept-the-move-constructor-and-move-assignment-operator/)

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -7591,7 +7591,7 @@ PUGI__NS_BEGIN
 			_root_size = state._root_size;
 		}
 
-		void release()
+		void release() PUGIXML_NOEXCEPT
 		{
 			xpath_memory_block* cur = _root;
 			assert(cur);
@@ -8613,13 +8613,13 @@ PUGI__NS_BEGIN
 		}
 	}
 
-	template <typename T> PUGI__FN void delete_xpath_variable(T* var)
+	template <typename T> PUGI__FN void delete_xpath_variable(T* var) PUGIXML_NOEXCEPT
 	{
 		var->~T();
 		xml_memory::deallocate(var);
 	}
 
-	PUGI__FN void delete_xpath_variable(xpath_value_type type, xpath_variable* var)
+	PUGI__FN void delete_xpath_variable(xpath_value_type type, xpath_variable* var) PUGIXML_NOEXCEPT
 	{
 		switch (type)
 		{
@@ -11872,7 +11872,7 @@ PUGI__NS_BEGIN
 			return new (memory) xpath_query_impl();
 		}
 
-		static void destroy(xpath_query_impl* impl)
+		static void destroy(xpath_query_impl* impl) PUGIXML_NOEXCEPT
 		{
 			// free all allocated pages
 			impl->alloc.release();
@@ -12040,7 +12040,7 @@ namespace pugi
 	}
 
 #ifdef PUGIXML_HAS_MOVE
-	PUGI__FN void xpath_node_set::_move(xpath_node_set& rhs)
+	PUGI__FN void xpath_node_set::_move(xpath_node_set& rhs) PUGIXML_NOEXCEPT
 	{
 		_type = rhs._type;
 		_storage = rhs._storage;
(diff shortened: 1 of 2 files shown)


zeux:
Yeah I considered something like this before. I'm not sure about marking the delete paths as noexcept - I believe starting in C++17 noexcept is part of the function type. This - as far as I understand - means that deallocation callback will have to be marked as noexcept, which will be a breaking change.

brandl-muc:
noexcept cannot be part of typedefs so you cannot mark it as noexcept. However it would still be a bad idea to use something that throws as it is used in destructors which are implicitly noexcept. (and generally should not throw)

zeux:
Right - I'm not worried about memory deallocation functions throwing. I'm worried about C++17 making the compiler aware of the distinction between noexcept function pointer and non-noexcept function pointer, and generating a warning/error when used to deallocate memory.

brandl-muc:
Where is the problem if you leave the deallocation function pointer  `noexcept(false)` as it is already? It is perfectly legal to call a `noexcept(false)` function from a `noexcept(true)` one.
Then both types of functions can bind to the pointer. I don't expect compilers to warn in the above case because there would be a lot of false positives. E.g., on [coliru](http://coliru.stacked-crooked.com/a/437d3711a502c0f2) no warning is issued.

zeux:
> It is perfectly legal to call a noexcept(false) function from a noexcept(true) one.

I expected this to not be the case, based on your changes to `delete_xpath_variable` but I've never used `noexcept` before so I could be wrong. I'll need to look into this a bit.

brandl-muc:
Ah ok sure, I understand.
I googled a bit and found this SO question: [Can a noexcept function still call a function that throws in C++17?](https://stackoverflow.com/q/39763401/1969455)

Also we could discuss to what extent you would want to propagate `noexcept`. I only added it to the move special members and the functions they are calling. I assumed the deallocation function pointer to be "conceptually `noexcept`" and therefore stopped with functions that are calling `xml_memory::deallocate()`. It is also possible to check all functions and add `noexcept` if possible, or only add it to dtors to make them explicitly `noexcept`.

As a side node, obviously it would be nice to have `noexcept` move special members for `xml_document` which does not seem feasible with the current implementation. Is opening an issue to discuss this a good idea?

zeux:
> As a side node, obviously it would be nice to have noexcept move special members for xml_document which does not seem feasible with the current implementation. Is opening an issue to discuss this a good idea?

The current implementation can be noexcept unless PUGIXML_COMPACT is defined; if it is, in some cases moving has to allocate which could throw. This is currently unavoidable; fixing that requires some changes in compact encoding that I was somewhat reluctant to make for the sake of move.

From what I can see, gcc doesn't actually check control flow for noexcept function to issue warnings which is somewhat odd (except for Wnoexcept which does something else); https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61372 was filed a while ago with no apparent action. One possibility is to either leave the change as is, or keep the noexcept annotations on just move machinery (ctor, operator=, _move). If warnings come up in the future we can deal with them then I guess.

brandl-muc:
OK, I'll reduce the `noexcept`s in this PR.

I'm thinking of creating a second PR which contains the deallocation in a `try catch(...)` block. Throwing and catching is allowed in noexcept functions, but the exception [must not leave the function](https://stackoverflow.com/a/28975292/1969455).

I 'm also considering to create a third one with conditional `noexcept` for xml_document. What do you think?

brandl-muc:
OK, I'll reduce the `noexcept`s in this PR.

I'm thinking of creating a second PR which contains the deallocation in a `try catch(...)` block. Throwing and catching is allowed in noexcept functions, but the exception [must not leave the function](https://stackoverflow.com/a/28975292/1969455).

I 'm also considering to create a third one with conditional `noexcept` for xml_document. What do you think?

zeux:
Thanks! Merged - I think this is a good step. I don't think try/catch is a good idea - if the deallocation function does throw, I'm not sure what the exception handler would do. If an exception leaves the function, I'd expect std::terminate to get called which is reasonable - it's a fatal error.

noexcept for xml_document move functions would make sense. Since it's not available in some configurations it might make sense to create a special macro for this to keep the usage cleaner, like PUGIXML_NOEXCEPT_DOCMOVE or something.

PR #184, opened 2018-02-27
Enables usage of override specifier for MSVC compilers

brandl-muc:
MS won't be able to increase __cplusplus soon so as a workaround this pull request adds special treatment of the MSVC compiler. (similar to the support of move semantics)
The override specifier is [supported with compiler 17.0](http://en.cppreference.com/w/cpp/compiler_support) which is part of Visual Studio 2012.


zeux:
Thanks!

PR #185, opened 2018-03-02
Adds noexcept specifiers to the move special members of xml_document,…

brandl-muc:
… but only #ifndef PUGIXML_COMPACT

As discussed in #183 but I named the macro PUGIXML_NOEXCEPT_IF_NOT_COMPACT.
Originally I wanted to go with something like you proposed but that name might give wrong hints to a reader who only looks at the move special members and not an the macro definition. The current name clearly expresses when the member is `noexcept` and when it is not.

Also I would have liked to use the `noexcept(static_condition)` syntax, but it is not possible to use it with a macro without [uglyl hacks](https://stackoverflow.com/a/41268349/1969455).

### DIFF
--- a/src/pugixml.hpp
+++ b/src/pugixml.hpp
@@ -92,7 +92,7 @@
 #ifdef PUGIXML_HAS_NOEXCEPT
 #	define PUGIXML_NOEXCEPT noexcept
 #	ifdef PUGIXML_COMPACT
-#		define PUGIXML_NOEXCEPT_IF_NOT_COMPACT noexcept(false)
+#		define PUGIXML_NOEXCEPT_IF_NOT_COMPACT
 #	else
 #		define PUGIXML_NOEXCEPT_IF_NOT_COMPACT noexcept
 #	endif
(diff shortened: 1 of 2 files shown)


src/pugixml.hpp
```diff
@@ -91,8 +91,14 @@
 #endif
 #ifdef PUGIXML_HAS_NOEXCEPT
 #	define PUGIXML_NOEXCEPT noexcept
+#	ifdef PUGIXML_COMPACT
+#		define PUGIXML_NOEXCEPT_IF_NOT_COMPACT noexcept(false)
```
zeux: Is noexcept(false) necessary here, as opposed to just not specifying noexcept?

brandl-muc: It is not necessary, if you prefer omitting it then I'll change it.
I did it this way from a "better explicit than implict" POV.

zeux: Ok - please remove it if this is not necessary.

brandl-muc: Done

PR #189, opened 2018-03-10
Correct switch fallthrough warnings

limitedAtonement:
GCC 7.3.0 gave the following warning:
src/pugixml.cpp:10193:5: warning: this statement may fall through [-Wimplicit-fallthrough=]
without these attributes.

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -10267,9 +10267,9 @@ PUGI__NS_BEGIN
 
 				if (_rettype == xpath_type_boolean)
 					return _data.variable->get_boolean();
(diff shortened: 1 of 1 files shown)


zeux:
Which pugixml version were you building? This should have been fixed in #134 (using comments as opposed to attributes that have portability issues)

limitedAtonement:
Oh, I see. I had compiled against "latest" (which now is pointing to `v1.8`). Right, there were comments, but I see #134 adjusted them so that the compiler can see them properly. I'll use the code at master instead. Thanks!

PR #191, opened 2018-03-29
Suffixes for different build types

straubar:
I would like to suggest adding suffixes for debug (and other) builds. This is especially helpful in a multi-configuration environment such as Visual Studio on Windows. Currently, installing a different configuration build will always overwrite the previously installed one, making it hard for other projects in such a multi-configuration environment to properly include pugixml.

I'm however open to suggestions to resolve this in a different manner.


zeux:
This seems fine to me. Will CMAKE_CONFIGURATION_TYPES be defined for "normal" (single-configuration) CMake builds?

straubar:
I'm not 100% sure about that but found this here: [https://stackoverflow.com/a/31548693](https://stackoverflow.com/a/31548693)

The (now mostly deprecated) wiki says:
When using a multi-configuration generator, such as the one for Visual Studio, this variable contains a list of the available configurations.

In the documentation you can find:
Specifies the available build types on multi-config generators.
This specifies what build types (configurations) will be available such as Debug, Release, RelWithDebInfo etc. This has reasonable defaults on most platforms, but can be extended to provide other build types.

This clearly doesn't sound as if it was specifically prohibited for single-configuration systems to have it empty. But I doubt it will be set there, as it is ignored in such a system.

zeux:
Yeah - for end-to-end CMake builds this shouldn't matter; it might matter if CMake is used to build pugixml but not to consume the resulting binary, although it may not matter much. I was curios as to what exactly the problem is.

I want to wait on merging this PR before 1.9 gets released in case changing the names ends up being a problem; I'm still finalizing the specific changes that I want to get in for that so it might take a bit of time to merge this; hoping to get 1.9 out in April.

PR #193, opened 2018-04-09
cmake: always install the pkg-config file

eli-schwartz:
There's really never a reason to *not* want this installed. If an option
is needed to specify installing in a versioned subdirectory, this option
should be explicitly described rather than hidden in something else.

As an added bonus, this makes the CMake install code slightly *less*
complicated.


zeux:
Thanks!

PR #194, opened 2018-04-09
Split LICENSE to a separate file for easier distribution

ArchangeGabriel:
Also fix two http → https occurrences.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,24 @@
+This library is available to anybody free of charge, under the terms of MIT License:
+
+Copyright (c) 2006-2018 Arseny Kapoulkine
+
+Permission is hereby granted, free of charge, to any person
+obtaining a copy of this software and associated documentation
+files (the "Software"), to deal in the Software without
+restriction, including without limitation the rights to use,
+copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following
+conditions:
+
+The above copyright notice and this permission notice shall be
+included in all copies or substantial portions of the Software.
+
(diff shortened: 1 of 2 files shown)


zeux:
The http -> https corrections make sense. Can you clarify the case for having a LICENSE file separately? Does it have to be called LICENSE or can it be called LICENSE.md instead?

eli-schwartz:
It could be anything, but the motivation was that our distro packaging can install the LICENSE file rather than trying to e.g. `sed` the license contents out of the README, or install the whole README to `/usr/share/licenses/` (which would be kind of odd).

ArchangeGabriel:
The MIT license requires people to distribute the license file with the project in any form. As a distro packager, I have thus to provide this file alongside the rest of the pugixml package. However, providing the full README does not really make sense, especially with all the GitHub specific tags at the top. Hence, a separate file would be better for this purpose.

Regarding the file name, it can definitively LICENSE.md, or COPYING, COPYING.md, etc.

zeux:
Gotcha - makes sense. What I'm starting to do for some of my other projects is this:

- In LICENSE.md, I keep the license boilerplate + "MIT License" as the first line to make it clear. E.g. https://raw.githubusercontent.com/zeux/meshoptimizer/master/LICENSE.md

- In README.md, I have the following text:
> License
> This library is available to anybody free of charge, under the terms of MIT License (see LICENSE.md).

This makes it easier for people looking at the README to instantly know what the license is. Would that work?

ArchangeGabriel:
Would be perfect. Feel free to edit or discard my PR to make your changes. ;)

zeux:
This has been merged although GH lost the merge tracking because I rebased across the merge boundary :(

PR #212, opened 2018-06-01
Add Position-Independed Code option for static build

eagleivg:
Im try link statis library, but get message "recompile with -fPIC". This PR fix it.


zeux:
This change would set fPIC for many cases when you don't need it, which might make performance worse. Have you tried setting POSITION_INDEPENDENT_CODE attribute in CMake globally, like suggested here?
https://stackoverflow.com/questions/38296756/what-is-the-idiomatic-way-in-cmake-to-add-the-fpic-compiler-option

eagleivg:
yep, this solve my problem, thanks

PR #215, opened 2018-06-26
Use CMAKE_INSTALL_LIBDIR for pugixml.pc

c72578:
- Up to now, the libdir was hardcoded to "lib" inside pugixml.pc and
  the install directory of pugixml.pc was "lib/pkgconfig"
- Adds support for lib and lib64 by using CMAKE_INSTALL_LIBDIR variable


zeux:
Thanks!

PR #220, opened 2018-07-22
improvements in cmake build

emmenlau:
This PR adds minor improvements in the cmake based build. Th largest change is to remove the option `BUILD_SHARED_LIBS` because it is a cmake default. There is also no need to explicitly quantify `add_library()` with `BUILD_SHARED_LIBS` because cmake will automatically build shared or static based on `BUILD_SHARED_LIBS`.
All in all this PR should not break anything and slightly simplify the CMakeLists.txt. Is this useful and acceptable?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /build/
 /.vscode/
+.*.swp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,10 @@ if(NOT ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} STRLESS 3.1 AND ";${CMAKE_C
 	target_compile_features(pugixml PUBLIC cxx_long_long_type)
 endif()
 
-set_target_properties(pugixml PROPERTIES VERSION ${pugixml_VERSION} SOVERSION 1)
+set_target_properties(${PROJECT_NAME} PROPERTIES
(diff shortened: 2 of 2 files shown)


zeux:
This makes sense to me, modulo forcing `POSITION_INDEPENDENT_CODE ON` - static library builds shouldn't set this.

emmenlau:
The `POSITION_INDEPENDENT_CODE ON` will not hurt static libs, AFAIK it will just be ignored. But I can remove it if you prefer?

zeux:
POSITION_INDEPENDENT_CODE ON should not be ignored for static libs - there are valid reasons to use PIC for static libs (which is when a static lib is linked into a shared lib), but it has a small performance penalty so it's the wrong default.

I was trying to find the defaults for BUILD_SHARED_LIBS in CMake documentation and couldn't. I'm starting to lean towards keeping this a project-only setting, since that way all valid and supported options are clearly documented in CMakeLists file.

emmenlau:
I'm not an expert in this matter, but the cmake documentation specifies "The `POSITION_INDEPENDENT_CODE` property determines whether position independent executables or shared libraries will be created", so it is my understanding that it should not have an effect on static libraries (see https://cmake.org/cmake/help/v3.12/prop_tgt/POSITION_INDEPENDENT_CODE.html).

Should I check how to set a default for `BUILD_SHARED_LIBS`? I recall there was a way to query if a variable is already set or not.

PR #221, opened 2018-07-22
Symbol visibility

emmenlau:
This PR replaces the current manually defined PUGIXML_API PUGIXML_CLASS with the cmake auto-generated PUGIXML_EXPORT. This should work for correct symbol visibility on all platforms. It can be used also on Un*x in combination with `symbol visibility hidden`.

Honestly, I did not fully understand the current logic of PUGIXML_API and PUGIXML_CLASS, so maybe this PR misses something? Please let me know if you find it useful or if it requires improvements.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,25 +44,32 @@ else()
 endif()
 
 # Export symbols for shared library builds
-if(BUILD_SHARED_LIBS AND MSVC)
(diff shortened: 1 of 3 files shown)


zeux:
This would mean that pugixml has to be used with CMake only - which is a deal breaker. I'm not sure I understand the purpose of this PR, but as it stands it definitely will break existing workflows.

PR #227, opened 2018-08-05
Visual Studio Natvis visualization

WheretIB:
This PR adds custom visualization for pugixml types to improve debugging experience in Visual Studio.

All build mode switches expect for PUGIXML_COMPACT are covered. There will be a separate PR for compact mode.


zeux:
Looks like the only difference between wide and non-wide visualizers is the format specifier - `sb` vs `sub`. Just to confirm, what happens if there's no format specifier for these values? (e.g. `{value}` instead of `{value,sb}`) If that automatically uses the correct type that could mean we only need one file.

WheretIB:
`sb` and `sub` formats remove `""` and `L""` around the strings respectively.

It's mainly used to display a pretty preview of the node value depending on type:
Current view

Without that format specifier it will look like this:
Single file view

I can rework the display to require only one file with a view like:
Reworked

contrib/natvis/pugixml.natvis
```diff
@@ -0,0 +1,82 @@
...
+  </Type>
+  
+  <Type Name="pugi::xml_node_struct">
+    <DisplayString Condition="(header &amp; 0xf) == 0">node_null</DisplayString>
+    <DisplayString Condition="(header &amp; 0xf) == 1">node_document</DisplayString>
+    <DisplayString Condition="(header &amp; 0xf) == 2">node_element name={name,na}</DisplayString>
+    <DisplayString Condition="(header &amp; 0xf) == 3">node_pcdata value={value,na}</DisplayString>
+    <DisplayString Condition="(header &amp; 0xf) == 4">node_cdata value={value,na}</DisplayString>
+    <DisplayString Condition="(header &amp; 0xf) == 5">node_comment value={value,na}</DisplayString>
+    <DisplayString Condition="(header &amp; 0xf) == 6">node_pi name={name,na}</DisplayString>
```
zeux: node_pi should also display value (e.g. `<?include me?>`)

WheretIB:
I've replaced each node type specialization with a generic enum cast with conditional name and value display if they are available:
Conditional display

zeux:
Perfect, thanks!

PR #238, opened 2018-11-12
Update URL to https in pugixml.pc.in

c72578:


zeux:
Thanks! I'll need to change the URL everywhere else in the code/documentation...

Issue #236, opened 2018-10-23
Coverity scan reports four issues in pugixml code


cecilios:
Coverity scan reports four issues in pugixml code, all of medium impact. These are:

## Dereference after null check

Refers to code where a pointer or reference is checked against null or assigned null and then later dereferenced. Dereferencing a null pointer leads to undefined behavior, which is usually a program crash. Coverity reports this error in three cases:

- Checking against NULL and then dereferencing on a path on which it was null.
- Assigning NULL and then taking a path on which the value has not changed.
- Dereferencing the return value from dynamic_cast without first checking it against NULL. If you know that the value will always be non-null, then you can use static_cast to avoid a defect report.

Coverity reports two instances of this error in pugixml.cpp:

### 1. Issue in line 4691:

```
4679        PUGI__FN xml_parse_result load_buffer_impl(xml_document_struct* doc, xml_node_struct* root, void* contents, size_t size, unsigned int options, xml_encoding encoding, bool is_mutable, bool own, char_t** out_buffer)
4680        {
4681                // check input buffer
    1. Condition !contents, taking true branch.
    2. var_compare_op: Comparing contents to null implies that contents might be null.
    3. Condition size, taking false branch.
4682                if (!contents && size) return make_parse_result(status_io_error);
4683
4684                // get actual encoding
4685                xml_encoding buffer_encoding = impl::get_buffer_encoding(encoding, contents, size);
4686
4687                // get private buffer
4688                char_t* buffer = 0;
4689                size_t length = 0;
4690
    CID 117652 (#1 of 1): Dereference after null check (FORWARD_NULL)4. var_deref_model: Passing null pointer contents to convert_buffer, which dereferences it.
4691                if (!impl::convert_buffer(buffer, length, buffer_encoding, contents, size, is_mutable)) return impl::make_parse_result(status_out_of_memory);
4692
```

Details on how the null ptr is dereferenced:
```
2245        PUGI__FN bool convert_buffer(char_t*& out_buffer, size_t& out_length, xml_encoding encoding, const void* contents, size_t size, bool is_mutable)
2246        {
2247                // fast path: no conversion required
    1. Condition encoding == pugi::encoding_utf8, taking false branch.
2248                if (encoding == encoding_utf8)
2249                        return get_mutable_buffer(out_buffer, out_length, contents, size, is_mutable);
2250
2251                // source encoding is utf16
    2. Condition encoding == pugi::encoding_utf16_be, taking true branch.
[...]

PR #240, opened 2018-11-14
Coverity: code annotate false positives

c72578:
- Add code annotations according to:
  https://github.com/zeux/pugixml/issues/236

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -4688,6 +4688,7 @@ PUGI__NS_BEGIN
 		char_t* buffer = 0;
 		size_t length = 0;
 
+		// coverity[var_deref_model]
 		if (!impl::convert_buffer(buffer, length, buffer_encoding, contents, size, is_mutable)) return impl::make_parse_result(status_out_of_memory);
 
 		// delete original buffer if we performed a conversion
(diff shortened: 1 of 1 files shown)


src/pugixml.cpp
```diff
@@ -10863,6 +10865,7 @@ PUGI__NS_BEGIN
 			if (_next)
 				_next->optimize(alloc);
 
+			// coverity[var_deref_model]
```
zeux: Is this required given the change https://github.com/zeux/pugixml/commit/d9fadc74646171d72b9a539ecff7183de1228f00?

c72578: Commit https://github.com/zeux/pugixml/commit/d9fadc74646171d72b9a539ecff7183de1228f00 addressed the 4. issue in line 10926

c72578: Maybe @cecilios can have a look and test with Coverity again

cecilios: @c72578 Now quite bussy but I will check on this weekend.

cecilios:
I have checked with Coverity: It complains in two lines, 10928 and 10087. It seems that commit d9fadc7 didn't solve the issue in line 10928.
I will add a coverity annotation in 10928 and move the current one near 10087 and re-check

c72578:
Closed this PR in favor of https://github.com/zeux/pugixml/pull/241

PR #248, opened 2018-12-09
 tests: Add -fshort-wchar tests

zeux:
These tests are very tricky - in general -fshort-wchar is a dangerous
option because, since the standard library is not compiled with it, you
can't use any functions from either C or C++ standard library without
getting ABI mismatch.

The reason we want to use this is to do coverage testing on UTF32->UTF16
and UTF16->UTF32 conversion paths, that generally aren't hit on
gcc/clang. To do this, we carefully work around any internal calls to
wcslen/wcscmp that pugixml might be doing and don't use any wstring
functions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,10 @@ $(BUILD)/%.o: %
 	@mkdir -p $(dir $@)
 	$(CXX) $< $(CXXFLAGS) -c -MMD -MP -o $@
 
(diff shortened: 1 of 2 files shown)


zeux:
This might not be necessary if we can run MinGW builds on AppVeyor that should have -fshort-wchar and 16-bit wchar_t CRT.

zeux:
MinGW is now tested on AppVeyor and reporting code coverage results properly, which makes this unnecessary.

PR #251, opened 2018-12-10
pkg-config: Use CMake GnuInstallDirs FULL vars

mpoquet:
Hello,

This commit uses [GNUInstallDirs](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) absolute path variables instead of creating such paths manually.
- Use `CMAKE_INSTALL_FULL_LIBDIR` instead of `CMAKE_INSTALL_PREFIX}`/`CMAKE_INSTALL_LIBDIR`
- Use `CMAKE_INSTALL_FULL_INCLUDEDIR` instead of `CMAKE_INSTALL_PREFIX}`/`CMAKE_INSTALL_INCLUDEDIR`

This fixes an installation problem in [Nix](https://nixos.org/nix/) packages, as non-FULL variables are already absolute paths in this case.


zeux:
Thanks!

Issue #40, opened 2015-03-31
Some enhancements


operamint:
I have posted a working solution to issue #49 https://code.google.com/p/pugixml/issues/detail?id=49, on how to support std::string arguments without overloading each API function.

I also would suggest two simple shorthand functions in the xml_node class, which should not clutter the interface or pose any compability issues, AFAIK:

```
xml_node operator()(string_arg name) const { return child(name); }
xml_attribute operator[](string_arg name) const { return attribute(name); }
```

It is especially useful for deep nested xml files, where you want to pull out a few single values, e.g.:

```
id = doc.child("site").child("regions").child("africa").child("item").attribute("id").value();
```

can be written:

```
id = doc("site")("regions")("africa")("item")["id"].value(); 
```

operamint:
Actually, as the attributes are a unique set, the [] operator could work like the std::set operator, so that it will add a new attribute if it does not exist. Makes it very simple to set add attributes to elements, and to look up attributes that are mandatory. Optional attributes should be accessed with the usual attribute(name) method:

xml_attribute operator[](string_arg name) {
  xml_attribute att = attribute(name);
  return att ? att : append_attribute(name);
}

Example: 
  node["radius"] = 0.5; // add attribute
  node["radius"] = 1.0 // update it.
  ...
  double r = node["radius"].as_double(); // look up mandatory attribute

lexxmark:
This was the 1st thing I've added once cloned the library.

zeux:
FWIW there's a lot of different opinions on the syntax and semantics of the operators (should [] automatically add an attribute? should it only do it if the attribute is a non-const value?).

A straightforward answer is to only provide read-only operators for attributes, but that just saves a tiny bit of typing over explicit `.attribute` calls.

Overall it doesn't seem like it's an obviously good idea. It saves typing, which is great, but it introduces additional syntax that duplicates existing one - see #255 for a different take. I'm going to close this since I think it's best that pugixml core interface remains unambiguous and clear.

PR #255, opened 2019-01-11
Add xml_node xml_node::operator[](const char_t* name), more convenient for config updating

halx99:
Such as:
```cpp
pugi::xml_document doc;
doc.load_string("<cfg></cfg>");
auto cfg= doc.document_element();
cfg["name"].text().set("halx99");
cfg["age"].text().set(19);
cfg["sex"].text().set("male");
doc.save_file("demo1.xml");
```

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -5482,12 +5482,12 @@ namespace pugi
 		return (_root && _root->value) ? _root->value + 0 : PUGIXML_TEXT("");
 	}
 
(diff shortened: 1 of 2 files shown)


zeux:
Sorry, I don't think I can merge this. There are many different opinions on what the operator syntax should be like - for example, you can refer to #40 for a completely different set of operators with completely different meaning and semantics. I feel that the operators in general don't capture the intent well enough to be added to pugixml.

PR #257, opened 2019-01-28
make pugixml build static and shared via cmake

basti171:
We need to build static and shared libraries at the same time. This patch worked for us. I set building shared libs off, so its default behaviour is the same as before. The only difference is: If you only want to build shared libs, you have to set BUILD_SHARED_LIBS and unset BUILD_STATIC_LIBS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,6 @@ cmake_minimum_required(VERSION 3.0)
 
 project(pugixml VERSION 1.9)
 
-option(BUILD_SHARED_AND_STATIC_LIBS "Build both shared and static libraries" OFF)
 option(BUILD_SHARED_LIBS "Build shared instead of static library" OFF)
 option(BUILD_TESTS "Build tests" OFF)
 option(USE_VERSIONED_LIBDIR "Use a private subdirectory to install the headers and libs" OFF)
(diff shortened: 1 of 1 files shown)


basti171:
@zeux Please review. Thanks!

zeux:
I don't think I can merge this as is - this will break builds for people who build pugixml via depending on the CMakeLists.txt since the project name changed (and also since the required configuration for shared library building changed). I guess I would be up for something like BUILD_MULTIPLE_LIBS which renames both target to pugixmlshared & pugixmlstatic, although it complicates the CMakeLists.txt somewhat.

Out of curiosity, why do you need to build both builds as part of the same CMake invocation? Is that something that other projects commonly support as well? CMake doesn't seem to make this very easy...

basti171:
In cjson there is a BUILD_SHARED_AND_STATIC_LIBS
We use pugixml in buildroot which supports building static and shared libraries at the same time. We are using crossbuilding for arm where we use shared libs. For testing and tools we build statically linked to use the results on inhomogeneous X86 machines, so that we only have a libstdc++/libc requirements.

zeux:
Yeah - I'll be up for something like BUILD_SHARED_AND_STATIC_LIBS; this looks like it can seamlessly support your workflow and not break existing users. Mild preference over renaming both library targets in this mode, looks like cjson just adds a -static suffix, thus favoring shared library builds.

basti171:
That is probably a better solution. I am on it. I am not sure how pull requests are working here, this is my first contribution. I reverted that changes and pushed it to my branch, but this is included inside this pull request. Is it usual to fork again and close this or do you just cherry-pick after the reverted commit?

basti171:
I tested it via 
cmake  -DBUILD_SHARED_AND_STATIC_LIBS=ON     to build both
cmake -DBUILD_SHARED_LIBS=ON                             to build only shared
cmake   to build only static

The idea of my Patch is: make different targets for static and shared libraries. if BUILD_SHARED_AND_STATIC_LIBS is set, activate both. If BUILD_SHARED_LIBS is set use shared only and deactivate static. otherwise use static.
I am a little unhappy with duplicating target option. I did't find a solution for setting an option to a list of targets. The cjson CmakeLists.txt is a little bit stranger, because there is an extra BUILD_SHARED_AND_STATIC_LIBS evaluation for every option, which sets both targets.

zeux:
Would it be possible to keep the target names in case the option isn't enabled? The biggest compatibility issue is that normally you'd just include pugixml via CMake by doing `add_subdirectory`, and then add the dependency on pugixml target. It's because of this that I'd like to maintain the "pugixml" target name, and only use -shared/-static names when you need to build both at once.

zeux:
Oh, also the PR is fine - I can squash this before merging, so you can just push commits to the branch and the changes get picked up in this PR.

basti171:
I hope that is how I understood you. It's more similar to the cjson solution, except when compiling both libraries the targets contained static/shared suffix and remained pugixml if it is built as before. That should not break anything existing

zeux:
Thanks! This is what I had in mind - this should work for you and should keep all existing users happy. The duplication in the CMake files is unfortunate but I guess that's what it is. I'll merge it and double check - I'm wondering if it's possible to define a list variable that refers to either pugixml or a list (pugixml-static, pugixml-shared) and use it in most cases where it's necessary to operate on all targets.

PR #258, opened 2019-02-06
Refactor CMakeLists.txt support for multiple targets

zeux:
We now have a ${LIBRARY} variable that we can either use directly or in
a foreach loop to be able to process either pugixml or pugixml-static
and pugixml-shared targets.

Also fixes incorrect shared library assignment when
BUILD_SHARED_AND_STATIC_LIBS is defined, and only links the static
library in for make check.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,8 +39,14 @@ if(DEFINED BUILD_DEFINES)
 endif()
 
 if(BUILD_SHARED_AND_STATIC_LIBS)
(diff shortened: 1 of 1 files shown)


zeux:
@basti171 fyi hopefully this works for you as well - I've checked this with cmake/make on Linux and it seems to properly build either one or both targets as instructed.

basti171:
works for me. Thanks. I am not very familiar with cmake. I didn't know foreach, good to know

PR #263, opened 2019-03-07
format_skip_control_chars

### LINKED ISSUE #262: Form feed character is encoded as &#12; which is invalid xml 1.0
`\f` character after decoding xml tree to string becomes `&#12;`. Well, any character within range [1; 31] except some valid ones (`\t \n \r`) is encoded this way.

Yanpas:
Fixes #262

### DIFF
--- a/docs/manual.adoc
+++ b/docs/manual.adoc
@@ -1704,6 +1704,8 @@ These flags control the resulting tree contents:
 
 * [[format_no_empty_element_tags]]`format_no_empty_element_tags` determines if start/end tags should be output instead of empty element tags for empty elements (that is, elements with no children). This flag is *off* by default.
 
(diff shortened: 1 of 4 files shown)


zeux:
Thanks, the change itself looks good. It would be great to add a simple unit tests and add this flag to manual (docs/manual.adoc) - you can find any other format flag in tests/docs for guidance. No need to rebuild HTML documentation, the setup process for that is nontrivial if you’re on Windows...

Yanpas:
Done. By the way debugging unittest was tricky since it only printed expected string, but not the actual result.

And about 18 unit tests were failing on Linux gcc 8.2.0... I'll take a look while be near my linux machine, maybe the cause was I was running them from wrong directory.

zeux:
`make test` should run them from the correct folder

zeux:
Merged, thanks!

PR #265, opened 2019-03-14
Remove contrib folder

### LINKED ISSUE #264: Why is the file pugixml/contrib/foreach.hpp under Public domain and not under MIT ?
Hello,
why is the file  pugixml/contrib/foreach.hpp under Public Domain licensed and not under MIT like all other files of pugixml? 
Public Domain is not easy to use in Gemany due to specific Copyright laws.

Is there a possibility to get the pugixml/contrib/foreach.hpp file under the MIT license?

Thanks in advance.
Siegfried Oesterreicher

zeux:
This change removes contrib folder since it doesn't seem very useful;
Natvis scripts can and should go to scripts/ along with project files.

foreach.hpp is supporting a severely outdated BOOST_FOREACH construct;
in C++11 ranged for loop can be used with xml_node::children/attributes,
and BOOST_FOREACH can work with these as well.

foreach.hpp was also accidentally licensed as public domain which isn't
very helpful when the actual library is MIT. We could fix the license
but it seems better to just remove it.

Fixes #264.


zeux:
I will keep this PR open for a bit to make sure this isn't a problem for any pugixml users, chime in if you use contrib/foreach.hpp!

c72578:
The removal of the `contrib` folder should also be considered in the following files:
`Makefile`
`readme.txt`

zeux:
Good catch, thanks!

PR #270, opened 2019-04-25
add a string writer

inbei:


zeux:
Sorry, but this PR can't be merged in its current form. The implementation doesn't have tests or documentation, doesn't match existing coding style and is substantially slower and more error-prone than a string writer that already exists in example code (see `samples/save_custom_writer.cpp`). While these issues can likely be fixed, for now it makes most sense to close this.

Issue #274, opened 2019-06-07
Support single quotes for enclosing attribute values


m-naumann:
Hey @zeux ,

Thanks for your great work! Would you be willing to merge a PR for allowing either also single quotes `'` instead of double quotes `"` [for enclosing attribute values](https://github.com/zeux/pugixml/blob/master/src/pugixml.cpp#L4082) or a custom enquoting char?

That would be very handy for our [lanelet2](https://github.com/fzi-forschungszentrum-informatik/Lanelet2)-map-framework which allows to use [JOSM](https://josm.openstreetmap.de/) for map creating and editing, because JOSM uses single quotes in the map files, and this makes git-diffs always huge and untransparent.

Thanks

zeux:
Yup, for sure! This has been on the todo list for a while but absent a request it just didn’t become prioritized. If you want to make a PR, go ahead - please use a format flag for this to minimize API changes, e.g. format_attribute_single_quote

PR #275, opened 2019-06-07
Feature single quote support

m-naumann:
closes https://github.com/zeux/pugixml/issues/274

### DIFF
--- a/docs/manual.adoc
+++ b/docs/manual.adoc
@@ -1706,6 +1706,8 @@ These flags control the resulting tree contents:
 
 * [[format_skip_control_chars]]`format_skip_control_chars` enables skipping characters belonging to range [0; 32) instead of "&#xNN;" encoding. This flag is *off* by default.
 
+* [[format_attribute_single_quote]]`format_attribute_single_quote` enables using single quotes `'` instead of double quotes `"` for enclosing attribute values. This flag is *off* by default.
+
 These flags control the additional output information:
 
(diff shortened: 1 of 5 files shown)


zeux:
I believe this is missing a change to node_output for escaping - will need to add escaping using & apos; and extend ctc_special_attr mask to include single quote.

m-naumann:
I think you are talking about these lines: https://github.com/zeux/pugixml/blob/master/src/pugixml.cpp#L2530 and https://github.com/zeux/pugixml/blob/master/src/pugixml.cpp#L2570
where you replace `&apos;` by `'` and `&quot;` by `"`, right?

Can you explain again what to change there (and maybe why, for my understanding)? 
(The characters quote and apostrophe are still the same, right? We don't want to exchange them, do we?)

And I couldn't find `ctc_special_attr` in the source code, can you provide a line for that? (Again for my understanding: You were supporting reading single quotes already, it's just about writing them now, right?)

zeux:
> Can you explain again what to change there (and maybe why, for my understanding)?
> (The characters quote and apostrophe are still the same, right? We don't want to exchange them, do we?)

My point is that before your change, with double quotes used as an attribute character, this:

```
node.append_attribute("name").set_value("val\"ue");
```

would result in this XML output:

```
<node name="val&quot;ue" />
```

After your change, setting the flag to use single quotes and using this code:

```
node.append_attribute("name").set_value("val'ue");
```

would result in this XML output:

```
<node name='val'ue' />
```

which is malformed; the correct output is `'val&apos;ue'`.

> And I couldn't find ctc_special_attr in the source code, can you provide a line for that?

Sorry, I meant `ctx_special_attr`. `chartypeex_table` needs to be extended to mark single quote as "special" character in attributes, and `text_output_escaped` needs to be extended to escape single quote as well, not just double quote.

m-naumann:
Okay, got it, should be fixed by https://github.com/zeux/pugixml/pull/275/commits/bcfdfd59b1ac9f8571b12bc88cded903d15158c3

zeux:
Thanks!

PR #283, opened 2019-08-03
fix cmake option USE_POSTFIX

wilhelmtell:
Appending the suffix to the build product need not be a function of whether `CMAKE_CONFIGURATION_TYPES` is set. For example, having two ninja build trees—one for debug and another for release—is a fine use case for `USE_POSTFIX`.


zeux:
Thanks!

PR #295, opened 2019-09-16
Add LICENSE.md to RELEASE

c72578:


zeux:
These files seem redundant wrt readme.txt that's already included in the archive (which is why they were left out) - can you clarify why they are necessary?

c72578:
The file LICENSE.md looks like the obvious "license file" of pugixml. "License files" are frequently packaged together with the source. Furthermore, the file LICENSE.md is referenced from README.md:

> This library is available to anybody free of charge, under the terms of MIT License (see LICENSE.md).

So I guess it would make sense to distribute both, LICENSE.md for the license file and README.md for some initial reading/information.

zeux:
But readme.txt already includes initial reading  information as well as the full license text (and license boilerplate is also duplicated in headers/sources).

I guess my point is that not including this into the release archive is not an oversight, it’s a conscious choice to minimize the amount  of redundancy (which readme do you open? If readme.txt is omitted, what do novice users do with the fact that .md doesn’t have a default Windows file association? etc).

So including this requires that to solve some problem that’s more important than having two readme files in the archive.

hobbes1069:
From a packaging point of view it’s preferable to have the README and LICENSE separate. For low space devices (arm) the license file is installed by necessity but other documentation is not.

zeux:
I am probably missing some context here. If having a license file be separate from readme is a packaging requirement I am up for including LICENSE.md into the release archives but it doesn’t seem like this necessitates inclusion of README.md.

c72578:
OK, I have updated the PR now, so that only LICENSE.md is added to RELEASE.

hobbes1069:
It's not a hard requirement but is typical. Pretty much all of the 150 packages I maintain for Fedora have a LICENSE or COPYING file. It gets marked as a license file in the RPM spec file while the rest get marked as documentation.

zeux:
Thanks!

Issue #282, opened 2019-07-18
Removing all children of node with one function


NeoProgramming:
pugi::xml_node has not member function like "clear" or "remove_all" for removing all children subnodes. It is possible to remove only single child node (xml_node::remove_child()), but this method perform additional actions, such as relink sibling nodes. This actions are necessary for removing single child node, but they are not necessary for removing all children of node.
Is it possible to add 'clear' method to pugi::xml_node?

zeux:
Something like `xml_node::remove_children` and `xml_node::remove_attributes` would make sense to me.

tuananh:
@zeux 

should it be sth like this?

```cpp
PUGI__FN bool xml_node::remove_children()
	{
		for (pugi::xml_node child: this->children())
		{
			remove_child(child);
		}

		return true;
	}
```

zeux:
The implementation should work with internal APIs - probably will need to duplicate parts of `remove_child` - for efficiency; should have a loop through children using `xml_node_struct*`, and be careful to read the next sibling before the node gets destroyed - see `destroy_node` for an example. The implementation of remove_attributes would be similar.

zeux:
In fact the optimal implementation should only call `impl::destroy_node`, and then manually clear out `first_child` pointer - this avoids the redundant node relinking mentioned in the original post.

tuananh:
this is my attempt but it still fails. the child is still there.

```cpp
	PUGI__FN bool xml_node::remove_children()
	{
		for (pugi::xml_node child: this->children())
		{
			if (!_root || !child._root || child._root->parent != _root) return false;

			impl::xml_allocator& alloc = impl::get_allocator(_root);
			if (!alloc.reserve()) return false;

			impl::destroy_node(child._root, alloc);
			child._root->first_child = nullptr;
		}

		return true;
	}
```

test

```TEST_XML(dom_node_remove_children, "<node><n1/><n2/><n3/><child><n4/></child></node>")
{
	xml_node node = doc.child(STR("node"));
	xml_node child = node.child(STR("child"));

	child.print(std::cout);
	CHECK(child.remove_children());
	CHECK_NODE(child, STR("<child/></child>"));
}
```

tuananh:
ah it seems to be working now. i will continue to work on `remove_attributes`

PR #296, opened 2019-09-17
feat: add remove_attributes() and remove_children()

tuananh:
Address issue: #282

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -6059,6 +6059,23 @@ namespace pugi
 		return true;
 	}
 
+	PUGI__FN bool xml_node::remove_attributes()
+	{
+		for(pugi::xml_attribute a: this->attributes())
+		{
+			if (!_root || !a._attr) return false;
+			if (!impl::is_attribute_of(a._attr, _root)) return false;
+
+			impl::xml_allocator& alloc = impl::get_allocator(_root);
+			if (!alloc.reserve()) return false;
+
+			impl::destroy_attribute(a._attr, alloc);
+		}
+		this->_root->first_attribute = nullptr;
+
+		return true;
+	}
+
 	PUGI__FN bool xml_node::remove_child(const char_t* name_)
 	{
 		return remove_child(child(name_));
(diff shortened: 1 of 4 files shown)


src/pugixml.cpp
```diff
@@ -6077,6 +6094,22 @@ namespace pugi
 		return true;
 	}
 
+	PUGI__FN bool xml_node::remove_children()
```
zeux: This function will crash when xml_node is a null handle, this needs a null check for _root and a test that xml_node().remove_children() works.

src/pugixml.cpp
```diff
@@ -6077,6 +6094,22 @@ namespace pugi
...
+	{
+		for (xml_node_struct* child = this->_root->first_child; child; child = child->next_sibling)
+		{
+			if (!_root || !child || child->parent != _root) return false;
```
zeux: Why do we need this condition here?

src/pugixml.cpp
```diff
@@ -6077,6 +6094,22 @@ namespace pugi
...
+
+			impl::xml_allocator& alloc = impl::get_allocator(_root);
```
zeux: This will return the same object for all nodes, it should be hoisted outside of the loop (along with reserve)

src/pugixml.cpp
```diff
@@ -6077,6 +6094,22 @@ namespace pugi
...
+			if (!alloc.reserve()) return false;
+
+			impl::destroy_node(child, alloc);
```
zeux: This destroys the node, potentially deallocating storage and invalidating next_sibling. You have to read next_sibling *before* the loop. Please read the implementation of destroy_node, it has the loop that's exactly the same.

src/pugixml.cpp
```diff
@@ -6077,6 +6094,22 @@ namespace pugi
...
+		}
+		this->_root->first_child = 0;
```
zeux: No need to qualify _root with this

src/pugixml.cpp
```diff
@@ -6059,6 +6059,23 @@ namespace pugi
 		return true;
 	}
 
+	PUGI__FN bool xml_node::remove_attributes()
```
zeux: Note, I've left a few comments in remove_children, most/all of them apply here as well.

src/pugixml.cpp
tuananh: Initially, i copied everything the `remove_child` method. so this is not needed?

src/pugixml.cpp
```diff
@@ -6059,6 +6059,28 @@ namespace pugi
...
+	{
+		if (!_root) return false;
+		impl::xml_allocator& alloc = impl::get_allocator(_root);
+		if (!alloc.reserve()) return false;
+
+		for (xml_attribute_struct* attr = _root->first_attribute; attr;)
+		{
+			if (!attr) return false;
```
zeux: I believe these two checks are redundant (you will only ever see attributes of the node itself on the attribute list), but it’s also dangerous to return in the middle of this since some attributes were destroyed already.

zeux:
Thanks!

Issue #294, opened 2019-09-16
Missing VERSION_INFO when compiling as Windows DLL


Blonder:
The complete version info for the DLL e.g. Company Name, File Version, Program Name, Program Version a.s.o. is missing when compiling as Windows DLL. This leads to an empty Property Page with no information when showing the Windows File Properties Dialog of the DLL (right click onto the DLL and open 'Properties' dialog).
You can solve this by adding a new directory e.g. win32 to the src directory and add a pugixml.rc file to the new folder. This .rc file should include the VS_VERSION_INFO with CompanyName, FileDescription, FileVersion, LegalCopyright a.s.o.

PR #298, opened 2019-09-18
Add: Windows DLL version info header, this solves issue …

Blonder:
https://github.com/zeux/pugixml/issues/294


src/win32/pugixml.rc
```diff
@@ -0,0 +1,47 @@
+#include <winver.h>
+//#include "../pugixml.hpp"
+// TODO these defines should be moved to ../pugixml.hpp
```
zeux: Yeah I doubt the resource compiler can ever parse the full header. I think it's fine to leave the version numbers duplicated here, they are already spread across many many files and my usual process is to search for all occurrences of 1.9 and change them to 1.10, so this works :)

Blonder: 20190918_pugixml_windows_version_number

Blonder: We could add 
`# define PUGIXML_VERSION_MAJOR 1`
`# define PUGIXML_VERSION_MINOR 10`
`# define PUGIXML_VERSION_PATCH 0`
`# define PUGIXML_VERSION_NUMBER "1.10.0\0"`
to the `PUGIXML_VERSION` section and
`#ifndef RC_INVOKED` in the `pugixml.hpp` header for the rest of the header.

zeux: Ah I guess. I think it's fine like this.

PR #300, opened 2019-09-21
Custom precision

### LINKED ISSUE #285: customize double precision
see https://github.com/zeux/pugixml/issues/18#issuecomment-522169447
It can be set either globally or overridden locally (i.e. `xml_attribute::set_value` )

tuananh:
Attempt to fix #285


zeux:
This looks pretty good (will need to take a closer look later), it's missing documentation and xml_text::set_value support - but I also have one small but essential comment on the interface.

The precision argument should be on the new overload of set_value instead of the default argument on the existing overload, e.g.

```
bool set_value(float value);
bool set_value(float value, int precision);
```

The reason for this is pretty subtle and it has to do with ABI compatibility (pugixml is trying to be ABI compatible within the minor releases) - see https://www.acodersjourney.com/20-abi-breaking-changes/. This is important because pugixml ships as a package in many Linux distributions, and this would ensure that it can be updated safely without recompiling any dependent packages.

tests/test_dom_modify.cpp
```diff
@@ -1762,6 +1762,19 @@ TEST(dom_fp_roundtrip_min_max_double)
 	CHECK(fp_equal(node.text().as_double(), std::numeric_limits<double>::max()));
 }
 
+TEST(dom_fp_double_custom_precision)
+{
+	xml_document doc;
+	xml_node node = doc.append_child(STR("node"));
+	xml_attribute attr = node.append_attribute(STR("attr"));
+
+	attr.set_value(std::numeric_limits<double>::min(), 20);
```
zeux: Would be great to have a couple of tests that just test basic float values, e.g. set_value(1.0f, 5) and set_value(3.1415926f, 3).

src/pugixml.hpp
```diff
@@ -263,6 +263,9 @@ namespace pugi
 	// Nodes are indented depending on their depth in DOM tree, a default declaration is output if document has none.
 	const unsigned int format_default = format_indent;
 
+	const unsigned int default_double_precision = 17;
```
zeux: Type should be int here and below to match the argument type.

zeux:
Thanks!

PR #302, opened 2019-09-28
Modernize CMake

### LINKED ISSUE #301: Modernize CMake files
Hi! I'm a bit of a CMake expert and I've been going around to various projects that are well known in the C++ community and helping fix their `CMakeLists.txt` files so that they are neat, clean, *actually readable* (ish, because, hey its CMake) and meet modern requirements for CMake. I'm willing to do the heavy lifting here (and by that I mean all of it), but I was wondering if you have any desired requirements? Currently modern distros support much more recent versions, but even then using CMake 3.4 can be considered a "conservative" approach.

I'll be submitting a PR in the near future, but figured I'd open an issue for discussion so you don't get blindsided with a diff for your build files :)

slurps-mad-rips:
This will close #301 

As per @zeux's request, I've tried to keep the plethora of configuration options available under the current workflow. I've occasionally marked one or two areas where some improvements could be made. That said, please let me know if this appears to change the current workflow or might cause too much disruption with downstream projects. I'd like to keep these changes as unintrusive as possible.

### New Features:

Bump CMake minimum to 3.4

pugixml no longer requires a C compiler to be found or set to "configure" correctly. This speeds up configuration and building on windows, as fewer processes are executed.

pugixml-shared and pugixml-static are "always" available, but not always
built. This allows downstream projects to still use them, and to make
sure our configurations are correct regardless of if they are being
built. They are not always installed however, as this would otherwise cause many *many* problems.

pugixml-shared and pugixml-static have ALIAS libraries of pugixml::shared and pugixml::static respectively. These names are also respected in the `find_package(pugixml CONFIG)` generated files, so its safe to use pugixml in a CMake project regardless of whether it is installed locally, or if its a subproject via `add_subdirectory`.

pugixml will automatically select the correct library type based on `BUILD_SHARED_LIBS`. A pugixml::pugixml ALIAS is also available.

`CMAKE_MSVC_RUNTIME_LIBRARY` from CMake 3.15 has been backported, and can be configured via the `STATIC_CRT`

CMake will now rely on generator expressions (a powerful abstraction in large projects with many subdirectories) for most work. This offloads work from the single-threaded configure stage to the multithreaded generation stage.

pugixml now uses CTest as the runner.

Some settings are automatically disabled if pugixml is used as a subdirectory. These are still able to be manually set, but are hidden from folks who choose to use CMake GUI, and use pugixml as a subproject.


zeux:
A few comments:

1. This is pretty much a complete rewrite of the CMake file. It'll take me some time to get through that.
2. Would it be possible to move .cmake.in file to scripts/ folder as to not pollute the root?
3. As for MSVC runtime, it would be nicer if the "backporting" code was explicitly bracketed by a CMake version check - presumably if the user is running on CMake 3.15 most of this code isn't necessary?
4. Not a big fun of CTest. Are there specific gains to be had by using it? Maybe I'm not using it fight, but ctest doesn't build the test changes so any test modifications now require running cmake --build and ctest manually; the default output is somewhat verbose but it omits *any* output from the test suite, even when tests fail, so I get this:

```
C:\work\pugixml-cmake\cmake>ctest -C Debug                                                                                                                                                                                                                                           Test project C:/work/pugixml-cmake/cmake                                                                                                                                                                                                                                                 Start 1: pugixml::test                                                                                                                                                                                                                                                           1/1 Test #1: pugixml::test ....................***Failed    0.65 sec                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      0% tests passed, 1 tests failed out of 1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  Total Test time (real) =   0.67 sec                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       The following tests FAILED:                                                                                                                                                                                                                                                                    1 - pugixml::test (Failed)                                                                                                                                                                                                                                                 Errors while running CTest                                                                    
```

Compare this to the old-style setup that immediately builds & tells you what's wrong (although with VS2019 it looks like there's a bunch of noise after this):

```
C:\work\pugixml\cmake>cmake --build .                                                                                                                                                                                                                                                Microsoft (R) Build Engine version 16.3.0-preview-19456-02+ee8294b55 for .NET Framework                                                                                                                                                                                              Copyright (C) Microsoft Corporation. All rights reserved.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   pugixml.vcxproj -> C:\work\pugixml\cmake\Debug\pugixml.lib                                                                                                                                                                                                                           check.vcxproj -> C:\work\pugixml\cmake\Debug\check.exe                                                                                                                                                                                                                               Test xpath_xalan_axes_2 failed: "@*" does not evaluate to expected set in context center at C:\work\pugixml\tests\test_xpath_xalan_3.cpp:42                                                                                                                                          FAILURE: 1 out of 943 tests failed.
```

slurps-mad-rips:
> Would it be possible to move .cmake.in file to scripts/ folder as to not pollute the root?

Sure thing! I'll put it in my next commit

> As for MSVC runtime, it would be nicer if the "backporting" code was explicitly bracketed by a CMake version check - presumably if the user is running on CMake 3.15 most of this code isn't necessary?

Most users are on not-3.15 and probably will be for a very long time. IIRC, even Ubuntu 20.04 will not be using CMake 3.15 (but don't quote me on that 😅)

This code gets eliminated at generation time due to the `$<VERSION_LESS:${CMAKE_VERSION},3.15>` expression. Effectively, none of these options are applied if we're on 3.15, while also allowing 3.15's approach to continue to work. It's basically a graceful fallback. The reason we have so many calls to `set()` is because, frankly, generator expressions can get huge and unwieldly. The generation step is multithreaded, unlike the configure step, so putting all the logic in generator expressions gives better performance in most situations.

> Not a big fan of CTest. Are there specific gains to be had by using it? Maybe I'm not using it right, but ctest doesn't build the test changes.

I added a custom target called "check" that does this part for you. Simply add `--target check` to get it to build and run ctest (So, `cmake --build . --target check`). As for outputting on failure, I can make a slight adjustment so that output is actually dumped on failure. I'd recommend you use CTest as there's no way for me to build pugixml's tests and then selectively disable them via a regex, or only run them via a regex, otherwise due to the `POST_BUILD` command. I'm not a huge fan of CTest myself, however, lots of projects use it because its there. The nice thing is that if someone were to override CTest in a CMake library, they'd get all the tests from projects and could setup a custom runner in those cases. If CTest was not used, this would unfortunately be impossible.

zeux:
> Most users are on not-3.15 and probably will be for a very long time.

Right, but I am hoping that if the 3.15 fallback is inside a separate if(), it'l be more obvious what code is responsible for it, and easier to clean up once 3.15 is common place (10 years?..)

slurps-mad-rips:
@zeux Updated! If you think we should go a step further let me know :)

CMakeLists.txt
```diff
@@ -1,105 +1,215 @@
...
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+  add_dependencies(check pugixml-check)
+  target_link_libraries(pugixml-check
+    PRIVATE
+      $<IF:${build-both},pugixml::static,pugixml::pugixml>)
+endif()
 
-configure_file(scripts/pugixml.pc.in ${PROJECT_BINARY_DIR}/pugixml.pc @ONLY)
-install(FILES ${PROJECT_BINARY_DIR}/pugixml.pc DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}/pkgconfig)
+#[[
```
zeux: Presumably this can be removed?

slurps-mad-rips: I had moved them further up the file.  See lines 160 and 178

slurps-mad-rips: oh wait you're talking about the commented out code. haha sorry.

slurps-mad-rips: Yes it can. I'll put a small push up in a hot second.

zeux:
Alright thanks. This looks good to me - it's unfortunate that the CMake file grew to be 2x the size but I guess such is life.

I'm going to merge this - fyi for previous CMake contributors @mpoquet @c72578 @eli-schwartz who if I'm not mistaken work on Linux packaging - there's no way for me to make sure that this doesn't break Linux distribution builds but the good news is that a point version (1.10) of pugixml was just released *before* this change, so there's a lot of time to refine the CMake config to fix anything that's amiss.

slurps-mad-rips:
I'm more than willing with work with them to make sure it can still produce proper linux packages. That said, a lot of the growth is for backporting the CMake 3.15 behavior. All of that can be removed when 3.15 is a "guaranteed minimum" in the general ecosystem. No idea when that would be however 😅

CMakeLists.txt
```diff
@@ -1,116 +1,210 @@
...
+  NAMESPACE pugixml::
+  DESTINATION ${CMAKE_INSTALL_DATADIR}/cmake/pugixml)
+
+install(FILES
+  "${PROJECT_BINARY_DIR}/pugixml-config-version.cmake"
+  "${PROJECT_BINARY_DIR}/pugixml-config.cmake"
+  DESTINATION ${CMAKE_INSTALL_DATADIR}/cmake/pugixml)
+
+install(FILES ${PROJECT_BINARY_DIR}/pugixml.pc
+  DESTINATION ${CMAKE_INSTALL_DATADIR}/pkgconfig)
```
eli-schwartz: This seems wrong, the datadir is in /usr/share, whereas this used to be installed in /usr/lib.

Since it contains the libdir path and points to a .so, this needs to be installed to the libdir, not to the datadir.

This means that on Arch Linux (which I care about) this would be broken if we tried to create a lib32-pugixml package for multilib support.

Debian, which has native multi-arch, will definitely care about i386 and amd64 versions of the package installing the same file with different contents in /usr/share.

slurps-mad-rips: ahh damn. That's a copy paste error. My mistake. I've been editing like 4 projects worth of CMake files. (TBQH I had figured no one was writing lib32/lib64 combinations anymore, but that's what I get for my hubris)

I can submit another PR with this fixed.

eli-schwartz:
The only complaint that is immediately apparent to me is simply that you've used the wrong filesystem hierarchy directory to store the cmake/pkg-config files, so it won't conform well to some expectations.

As I'm no great expert on cmake myself, I'll sort of assume everything else is good. The pkg-config files are the only reason I've ever contributed to this project anyway. :p

(The good news is we most definitely have 3.15.4 available.)

slurps-mad-rips:
@eli-schwartz unfortunately, a lot of LTS linux distros have CMake 3.5 or similar so life is... not great at the moment for me as I try to fix everyone's CMake files 😅

slurps-mad-rips:
PR pushed at #303

PR #303, opened 2019-10-06
Fix install locations for pkgconfig *and* cmake

slurps-mad-rips:
Per @eli-schwartz's recommendations, this PR fixes the pkgconfig file locations. In an attempt to also future-proof it for CMake's config files, we now install them to libdir as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,10 +173,10 @@ install(EXPORT pugixml-targets
 install(FILES
   "${PROJECT_BINARY_DIR}/pugixml-config-version.cmake"
   "${PROJECT_BINARY_DIR}/pugixml-config.cmake"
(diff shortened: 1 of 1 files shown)


zeux:
@slurps-mad-rips I was reading the CMake files more closely - after this change, there are still two references to CMAKE_INSTALL_DATADIR. Is this intentional? Should they point to LIBDIR as well or not? The second in particular refers to `${CMAKE_INSTALL_DATADIR}/cmake/pugixml` and I'm not sure how this interacts with references to `${CMAKE_INSTALL_LIBDIR}/cmake/pugixml` elsewhere.

slurps-mad-rips:
Ahh damn. You're right. That will make it an issue. It needs to be changed to `CMAKE_INSTALL_LIBDIR` for those values as well. I'll have a PR up in a moment

PR #305, opened 2019-10-18
Fix incorrect installation location for exported CMake targets

slurps-mad-rips:
This fixes the issue discussed just mentioned in #303


zeux:
Thanks!

PR #310, opened 2019-11-27
explicit casts for -Wconversions warnings in gcc

amlucas:
The current implementation complains if -Wconversions is set in gcc


zeux:
Thanks!

PR #311, opened 2019-11-27
Use 'Explicit Specialization of Function Templates' to make serializing could be extend with custom type.

halx99:
… could extend with custom type

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -5234,68 +5234,6 @@ namespace pugi
 		return _attr;
 	}
 
-	PUGI__FN xml_attribute& xml_attribute::operator=(const char_t* rhs)
-	{
-		set_value(rhs);
-		return *this;
-	}
-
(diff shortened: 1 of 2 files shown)


zeux:
So this generally speaking seems like a reasonable idea, but I can't really merge this.

There are two reasons why.

1. This breaks source compatibility. Any code that relied on implicit conversions to any of the types accepted by the overloads will now stop compiling. Now, in some sense it's probably good that it will, because implicit conversions are dangerous, but pugixml has a strict source compatibility policy across minor releases, and in this case I'm somewhat reluctant to break this in a new major version if/when it happens (although I'll need to think about this further).

2. This breaks binary compatibility. The specializations are mangled differently, which means that recompiling libpugixml.so will break dependent binaries. This is important for various Linux distributions, and pugixml has a strict binary compatibility policy across minor releases.

One additional thought is that it would be nice to make it so that the type extension can be applied to text and attribute nodes simultaneously, whereas this approach would require re-specializing type support twice.

Issue #299, opened 2019-09-21
Automated Static Code Analysis Vulnerabilities


codeanalysis9876:
Memory Leak - At Line: 3450 of File: external/pugi/pugixml.cpp
Memory Leak - At Line: 3468 of File: external/pugi/pugixml.cpp
Memory Leak - At Line: 3476 of File: external/pugi/pugixml.cpp
Memory Leak - At Line: 3630 of File: external/pugi/pugixml.cpp
Memory Leak - At Line: 6450 of File: external/pugi/pugixml.cpp
Memory Leak - At Line: 6674 of File: external/pugi/pugixml.cpp
Memory Leak - At Line: 6746 of File: external/pugi/pugixml.cpp
Memory Leak - At Line: 9587 of File: external/pugi/pugixml.cpp
Memory Leak - At Line: 9912 of File: external/pugi/pugixml.cpp
Null Deference - At Line: 9091 of File: external/pugi/pugixml.cpp
Memory Leak - At Line: 9587 of File: external/pugi/pugixml.cpp
Memory Leak - At Line: 9912 of File: external/pugi/pugixml.cpp
Unchecked Return Value - At Line: 3416 of File: external/pugi/pugixml.cpp
Unchecked Return Value - At Line: 3418 of File: external/pugi/pugixml.cpp
Buffer Overflow - At Line: 2910 of File: external/pugi/pugixml.cpp
Buffer Overflow - At Line: 2918 of File: external/pugi/pugixml.cpp
Buffer Overflow - At Line: 2919 of File: external/pugi/pugixml.cpp
Buffer Overflow - At Line: 2927 of File: external/pugi/pugixml.cpp
Buffer Overflow - At Line: 2928 of File: external/pugi/pugixml.cpp
Buffer Overflow - At Line: 2929 of File: external/pugi/pugixml.cpp
Dead Code - At Line: 7964 of File: external/pugi/pugixml.cpp
Dead Code - At Line: 7702 of File: external/pugi/pugixml.cpp
Dead Code - At Line: 7714 of File: external/pugi/pugixml.cpp
Dead Code - At Line: 7738 of File: external/pugi/pugixml.cpp
Dead Code - At Line: 7746 of File: external/pugi/pugixml.cpp
Dead Code - At Line: 7754 of File: external/pugi/pugixml.cpp
Dead Code - At Line: 7782 of File: external/pugi/pugixml.cpp
Dead Code - At Line: 7807 of File: external/pugi/pugixml.cpp
[...]

PR #313, opened 2019-12-02
Fixing possible null pointer dereference

valeriyvan:


zeux:
Is there any more detail on how this happened? Is this a real null pointer dereference, or is it reported by a static analysis tool (if so, which one?). Asking because as far as I can tell it should not be possible to get into this situation although the function logic is a bit messy and probably needs cleanup.

valeriyvan:
Static analyser of Xcode. And it seems like exclusion path where null dereference might happen really exists.

zeux:
The static analyzer seems to think that .root() of a non-null node can return a null node, but this is false.

valeriyvan:
It's the latest Xcode, which is Version 11.2.1.
Please have a look into diagnostic.
Sure, all static analysers might be false positive.
But it shows very deep execution path where nil might appear.
Probably there's a bug on some levels up.

zeux:
After looking at the code more I'm 99% confident this is not a real bug. Static analyzers sometimes may find issues but my general experience is summarized here: https://github.com/zeux/pugixml/issues/299#issuecomment-535100653. Regardless, as I said, it would be nice to keep the code clean of static analysis issues - I can reproduce this on clang9 and I'll take a look at the other two issues.

PR #316, opened 2019-12-06
Add POSITION_INDEPENDENT_CODE option

BotellaA:
Allow to use the static library in other dynamic librairies. This is a standard cmake option https://cmake.org/cmake/help/v3.11/prop_tgt/POSITION_INDEPENDENT_CODE.html#prop_tgt:POSITION_INDEPENDENT_CODE


panquez:
Any news on this PR ?

zeux:
Sorry I forgot to comment on this. I was somewhat hesitant to merge this as is - there's no configuration option for this, so it affects codegen for every user, including those that don't care about linking static libraries into PIE executables.

But on further reflection, the difference is probably negligible especially in x64 builds that I'd expect to use rip-relative addressing? I'm not sure if this is common practice - I haven't seen this often before.

I'm going to merge this as is for now. @slurps-mad-rips if you have any comments / suggestions post-merge they are welcome.

slurps-mad-rips:
@zeux As you said, the difference is negligible (though there are some reports that PIE improves support on GNU toolchains and doesn't affect the clang or msvc toolchains. Third party toolchains are up in the air, if I'm being honest).

I personally don't see an issue with it, however I'd have gated it behind the [`check_pie_supported`][1] command and then used `POSITION_INDEPENDENT_CODE ${CMAKE_CXX_LINK_PIE_SUPPORTED}`, as newer versions of CMake combined with old gnu toolchains could fail to build. That said, setting it to 'on' for the static library *could* theoretically be done downstream by users if they aren't installed pugixml onto their systems, but using it as part of a super build (either through `ExternalProject` or `FetchContent`). Additionally older versions of CMake don't understand PIEs at all (support was added in 3.13), so this could cause a few more bug reports in the future. That said the check to handle this *could* be done to some degree like:

```cmake
include(CheckPIESupported OPTIONAL)
if (POLICY CMP0083)
  cmake_policy(SET CMP0083 NEW)
endif()
if (COMMAND check_pie_supported AND NOT DEFINED CMAKE_CXX_LINK_PIE_SUPPORTED)
  check_pie_supported(LANGUAGES CXX)
endif()
set_property(TARGET pugixml-static
  PROPERTY
    POSITION_INDEPENDENT_CODE ${CMAKE_CXX_LINK_PIE_SUPPORTED})
```

I know this is a lot, but because we have minimum versions and CMake is "nightmare mode" it's the best option. The nice thing about this is that if someone calls `check_pie_supported` "above" us in a build hierarchy, we don't do the check and just use whatever was the result of their lookup.

It might make sense to also gatekeep this behind a `cmake_dependent_option`. I'm open to any of your opinions, @zeux. I've got the next two weeks off for holidays and I'll be tackling a few open issues that have cropped up across multiple open source projects (because "I care") and can add pugixml to the list if you think this sort of change is warranted.

[1]: https://cmake.org/cmake/help/latest/module/CheckPIESupported.html

slurps-mad-rips:
also, as is typical with cmake, what seems to be easy rarely is in the long run 😔

Issue #320, opened 2020-01-12
invalid character escaping for pcdata


pamapa:
According to https://www.w3.org/TR/xml11/#sec-predefined-ent these characters must be handled special. Also see here: http://www.java2s.com/Tutorial/XML/0020__Introduction/IllegalPCDATACharactersandEscapingCharacters.htm

```
&amp;    - the & character 
&lt;     - the < character 
&gt;     - the > character 
&apos;   - the ` character 
&quot;   - the " character
```
The encoding must also apply for PCDATA

pamapa:
dropped first pull request, as commit f9bd700 was wrong and made a new pull request

zeux:
> The encoding must also apply for PCDATA

I don't believe this is correct. The section you linked says that the references must be unescaped if present - which is correct - but XML specification doesn't require escaping '/" in PCDATA.

https://www.w3.org/TR/xml11/#syntax says:

> The ampersand character (&) and the left angle bracket (<) must not appear in their literal form, except when used as markup delimiters, or within a comment, a processing instruction, or a CDATA section. If they are needed elsewhere, they must be escaped using either numeric character references or the strings "&amp;" and "&lt;" respectively. The right angle bracket (>) may be represented using the string "&gt;", and must, for compatibility, be escaped using either "&gt;" or a character reference when it appears in the string "]]>" in content, when that string is not marking the end of a CDATA section.

With a following BNF form:

> [14] | CharData | ::= | [^<&]* - ([^<&]* ']]>' [^<&]*)

Both of these suggest that ampersand and left bracket must be escaped, but none of these suggest that quotes must be.

pamapa:
Thanks for your fast answer. You are correct. I was confused, sorry. I also double checked with a wide spread XML handling implementation (python's xml.etree.cElementTree):

```
import xml.etree.cElementTree as ET

root = ET.Element("root")
doc = ET.SubElement(root, "doc")
ET.SubElement(doc, "node").text = "1'2\"3<4>5"

tree = ET.ElementTree(root)
tree.write("test_pcdata_python.xml")
``` 
Output
``` 
<root><doc><node>1'2"3&lt;4&gt;5</node></doc></root>
``` 

But i am still having a problem with a UPnP device. I will need to debug more into this. Would a special format option acceptable, if needed?

zeux:
> But i am still having a problem with a UPnP device. I will need to debug more into this. Would a special format option acceptable, if needed?

[...]

PR #322, opened 2020-01-12
fix: #320: " and ' encoding must also apply for PCDATA

pamapa:


pamapa:
" and ' is allowed in pcdata, no need to merge this one

PR #329, opened 2020-02-07
Add vcpkg installation instructions

PhoebeHui:
pugixml is available as a port in VCPKG , documenting the install process here will help users get started by providing a single set of commands to build pugixml , ready to be included in their projects.

VCPKG is a C++ library manager that simplifies installation for pugixml and other project dependencies, we also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,19 @@ Documentation for the current release of pugixml is available on-line as two sep
 
 You’re advised to start with the quick-start guide; however, many important library features are either not described in it at all or only mentioned briefly; if you require more information you should read the complete manual.
 
+## Building pugixml with VCPKG
+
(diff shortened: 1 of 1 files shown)


zeux:
Thanks for opening this! It's great to know that vcpkg is available as a way to build pugixml. However, I don't think it should be in the README, at least not like this.

The instructions you posted don't work on Windows and for novice users they aren't going to be super easy to understand. Since there's *no* other instructions in the README about building, it seems odd to single out vcpkg.

What about adding this to the manual instead to this section https://pugixml.org/docs/manual.html#install.getting ? It would be great if the bootstrap instructions include Windows setup as well.

PR #330, opened 2020-02-07
Update CMakeLists.txt to rename the static and shared libraries to pugixml

LocutusOfBorg:
Don't call them pugixml-shared or pugixml-static, because this breaks pkgconfig calls and userspace tools.


zeux:
Thanks!

Issue #342, opened 2020-04-18
umlaute in pcdata


pamapa:
Is there a way to special encode umlaute (öäü, ...) in pcdata when printing?
```
ö -> &#246;
```

Something like python is doing:
```
import xml.etree.cElementTree as ET
root = ET.Element("root")
doc = ET.SubElement(root, "doc")
ET.SubElement(doc, "node").text = "Hello Wö"
tree = ET.ElementTree(root)
tree.write("test_pcdata_python.xml")
```
gives:
```
<root><doc><node>Hello W&#246;</node></doc></root>
```

With pugixml gives (flags=0, encoding=auto):
```
...
x.append_child(pugi::node_pcdata).set_value(tmp.c_str());
doc.print(buf, "", 0);
```

```
<root><doc><node>Hello Wö</node></doc></root>
```

pamapa:
Would be cool to have a special flag to do something like this:

```
std::string xmlEncodePcData(const std::string& pcdata)
{
    std::ostringstream buf;
    for (size_t i = 0; i < pcdata.length();) {
        auto ch = pcdata[i];
        int cplen = 1;
        if ((ch & 0xf8) == 0xf0) cplen = 4;
        else if ((ch & 0xf0) == 0xe0) cplen = 3;
        else if ((ch & 0xe0) == 0xc0) cplen = 2;
        if ((i + cplen) > pcdata.length()) cplen = 1;

        if (cplen == 1) {
            switch (ch) {
                case '&' : buf << "&amp;"; break;
                case '<' : buf << "&lt;"; break;
                case '>' : buf << "&gt;"; break;
                default:
                    // handle control codes
                    if (((ch >= 0x00) && (ch <= 0x1f) && 
                         (ch != 0x09) && (ch != 0x0d) && 
                         (ch != 0x0a)) || (ch == 0x7f)) {
                            buf << '.';
                    } else
                        buf << ch;
                    break;
            }
        } else {
            unsigned int res = static_cast<unsigned char>(ch & (0xff >> (cplen + 1))) << ((cplen - 1) * 6);

            size_t j;
            int len;
            for (j = 1, len = cplen -1; len; --len) {
                res |= (static_cast<unsigned char>(pcdata[i+j]) - 0x80) << ((len - 1) * 6);
            }
            buf << "&#";
            buf << res;
            buf << ";";
        }

        i += cplen;
    }

    std::string encoded = buf.str();
    return encoded;
}
```

pamapa:
[...]

PR #343, opened 2020-04-18
#342 escape non-ascii characters too

pamapa:
see #342 

Things to discuss first:
- I am not sure sure with the name of the flag format_escape_nonascii
- Should format_escape_nonascii be on by default and we add a no flag? Other xml libs (e.g. Python's xml.etree.cElementTree) are doing this.

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -3903,53 +3903,114 @@ PUGI__NS_BEGIN
 		xml_encoding encoding;
 	};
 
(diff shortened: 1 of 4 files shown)


pamapa:
the failures do not seem anything to do with my changes. They are compilation warnings in other areas which are treated as errors.

zeux:
Closing for now, we can come back to this if it resurfaces (see linked issue for more details).

PR #350, opened 2020-05-04
clang-tidy fixes

neheb:


neheb:
Removed problematic commits and added another one.

neheb:
Added extra fixes.

neheb:
Removed problematic stuff. Travis seems totally broken.

zeux:
Closing this:

- NULL changes tracked here #586 (only interested in fixing gcc compilation warning, not clang-tidy)
- I will not merge C++ header changes because that results in needless compatibility complications for very old toolchains that I would still like to support
- Other stylistic fixes (like namespace comments and variable renames) should be fine but they will need to be submitted separately.

PR #360, opened 2020-07-06
[Cmake] Static build should not export APIs

### LINKED ISSUE #359: Wrong DLLExport declaration for static builds
Hi,

I'm building your library as a static one, but I noticed all of your APIs are declared as `dllexport` which actually is not required ( see https://github.com/zeux/pugixml/blob/master/CMakeLists.txt#L105 ).

I think you need to patch that line to something like `$<$<CXX_COMPILER_ID:MSVC>:PUGIXML_API=__fastcall>)`.

Any better suggestion is welcome, as long as the library does not export when built statically :)

Thank you in advance!

//EDIT: I did try a personal patch by removing entirely that block and it works just fine as well. Using MSVC latest with `CMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded`

julianxhokaxhiu:
Fixes #359

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,9 +100,6 @@ if (NOT BUILD_SHARED_LIBS OR BUILD_SHARED_AND_STATIC_LIBS)
   target_include_directories(pugixml-static
     PUBLIC
       $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>)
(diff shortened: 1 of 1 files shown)


zeux:
Thanks!

julianxhokaxhiu:
Thank you for merging it :)

PR #363, opened 2020-07-15
Fix debug postfix in pkgconfig file

c72578:
In case of `USE_POSTFIX`, the `POSTFIX` is dependent
on the `CMAKE_BUILD_TYPE`.
Use the correct `POSTFIX` also in the generated **`pugixml.pc`** file.

This results in the following contents of **`pugixml.pc`**:
- Release:
`  Libs: -L${libdir} -lpugixml`
- RelWithDebInfo
`  Libs: -L${libdir} -lpugixml_r`
- MinSizeRel:
`  Libs: -L${libdir} -lpugixml_m`
- Debug:
`  Libs: -L${libdir} -lpugixml_d`


zeux:
Thanks!

PR #364, opened 2020-07-19
[WIP] Fixed: MSVC warnings

TheNicker:


TheNicker:
It seems like this PR exposes weakness in the code base, now that I've added a ~~proper~~ better exception handing for VC14 and above.
the following line tries to open the file: 'tests/data/unknown/output.xml' but `unknown` folder cannot be magically created. https://github.com/zeux/pugixml/blob/master/tests/test_document.cpp#L777

Please review the aforementioned.

zeux:
The test you mention specifically checks error handling behavior for save_file.

Here's what I don't understand:

1. load_file and save_file don't throw exceptions. It's not clear to me why your change changes that?
2. I'm a bit confused as to why this change is necessary in general, seeing how pugixml silences the 4996 warning that this change is presumably trying to avoid by using pragma warning disable at the top of the file.

TheNicker:
In that case there's no logic error :thumbsup:, just a test case that needs to be amended.

1. While browsing the code I've noticed there's an optional support for CPP exceptions , so the question is why not ? 
2. The warning occurs in Clang for Visual studio, so `-Wdeprecated` probably doesn't work in this specific use case. Is deprecation considered by you as non-significant ? thus the concealment ?

zeux:
> While browsing the code I've noticed there's an optional support for CPP exceptions , so the question is why not ?

Two reasons:

- This change will break backwards compatibility as evidenced by the tests failing.
- The only place in the API where exceptions are thrown is XPath query parsing; this is an unfortunate historical artifact, and if there was a clean way to fix this by not throwing the exception without breaking existing users, I'd happily do that. (this might happen in 2.0)

> The warning occurs in Clang for Visual studio

Ah, ok, that clarifies things. It would be good to fix this, two possible solutions are silencing deprecating warnings in clang builds, or using _s functions *without changing error handling behavior*. `FILE* f = fopen(...)` should be more or less equivalent to `FILE* f; fopen_s(&f, ...);`

The deprecation of these specific functions is a MS specific problem and I don't perceive it as significant - when MS decides to call standard C/C++ functions deprecated and introduce custom variants that aren't well supported elsewhere and aren't more secure in this case (fopen isn't any worse than fopen_s in actual practice), conforming with this just reduces code portability. MS provides a way to disable these deprecations via _CRT_SECURE_NO_WARNINGS define but that doesn't work for header-only builds as it must be enabled project-wide in this case, unfortunately.

TheNicker:
I would expect the exception mechanism to be consistent across the library. It might be not too critical at this point but I would suggest adding it to the pipeline, even as an open issue for when breaking changes would be scheduled. 

------------

`fopen_s` is standard C (11) function, though the wide char version is indeed a Microsoft extension.

But whether it is better or worse is irrelevant. If the project has intention to continue support for Win32 platform, deprecated functions should not be used, rather sooner than later. 


Several points:
*  IMHO _CRT_SECURE_NO_WARNINGS  will affect less and less deprecated methods till this flag will have minimal or no use.

* `fopen` is a part of the standard C, so there's less likelihood that it'll be removed, though there's a chance that Microsoft will push for its removal and/or they will decide to remove it in their implementation.

* `_wfopen` is already a Microsoft extension so I would consider changing it to `_wfopen_s` ASAP.

PR #373, opened 2020-10-13
Simple recursive tree search

KarenHeart-DePaul:
Reads all of the tags and enclosed data within an XML file using a simple recursive approach. It demonstrates the use of direct recursion in this pugixml library, as well as how the content of different types of tags can be accessed.


KarenHeart-DePaul:
Added documentation of the traversal tree in the manual

PR #381, opened 2020-11-23
format-option to omit the whitespace inserted before the closing slash

ottigeda:
I would like to be able to produce the same output as my old parser did.
Therefore this format-option would be handy for me.


ottigeda:
Is there anything I can/should do, so someone does have a look at this pull-request and tell me your opinion ?
Maybe I missed some rules on how to contribute, I was not able to find.
If it is just that you did not find the time yet then ignore this comment and I wait :-)

zeux:
I'd like to keep this PR open for a while to see if this is a feature other people might need. There's nothing problematic with this change per se but there's already a *loooot* of format flags, and this one is for a very small change in the output format that you can also get with format_raw (I understand that you'd lose indentation that way) so I'm not comfortable with just blindly merging this unless there's a larger need.

A minor note but the comment for the option seems to be copy & pasted.

ottigeda:
OK, makes sense. 
Yes that would be exactly what I like to achieve: keeping the indentation, but remove the whitespace before the closing slash.
BTW: I updated the comment.

PR #384, opened 2020-12-02
Update 1.11 release notes

mathstuf:
---
This change came as news to us. Adding a release note and compatibility for others in the future :) .

### DIFF
--- a/docs/manual.adoc
+++ b/docs/manual.adoc
@@ -2156,6 +2156,9 @@ Maintenance release. Changes:
 	. Fix Wzero-as-null-pointer-constant warnings in pugixml.hpp
 	. Work around several static analysis false positives
 
+* Build system changes
+	. The CMake package for pugixml now provides a `pugixml::pugixml` target rather than a `pugixml` target. A compatibility `pugixml` target is provided if at least version 1.11 is not requested.
+
 [[v1.10]]
 === v1.10 ^2019-09-15^
 
--- a/scripts/pugixml-config.cmake.in
+++ b/scripts/pugixml-config.cmake.in
@@ -1,3 +1,10 @@
 @PACKAGE_INIT@
 
 include("${CMAKE_CURRENT_LIST_DIR}/pugixml-targets.cmake")
+
+# If the user is not requiring 1.11 (either by explicitly requesting an older
+# version or not requesting one at all), provide the old imported target name
+# for compatibility.
+if (NOT DEFINED PACKAGE_FIND_VERSION OR PACKAGE_FIND_VERSION VERSION_LESS "1.11")
+  add_library(pugixml ALIAS pugixml::pugixml)
+endif ()


zeux:
This definitely wasn't intended. CMakeLists.txt does have this:

```
add_library(pugixml INTERFACE)
target_link_libraries(pugixml INTERFACE ${pugixml-alias})
add_library(pugixml::pugixml ALIAS pugixml)
```

So I'd think that pugixml::pugixml and pugixml are already equivalent - is that not the case?

zeux:
(and your change would make the alias self-recursive which doesn't seem right, but I'm not a CMake expert so maybe I'm missing something)

mathstuf:
My change is for users of `find_package(pugixml)` which work with 1.10 so that they continue to work with 1.12 and newer (1.11 being the "broken" release without compat).

zeux:
Mhm - before the CMake rework I can see this worked as is after using find_package:

> target_link_libraries(test pugixml)

But after it doing this doesn't seem to add public includes so `#include "pugixml.hpp"` doesn't work. Using `pugixml::pugixml` target fixes this.

I'd rather fix this unconditionally because the intention was to make it possible to depend on `pugixml` throughout the transition but I'm not sure what the best option to do so is - a circular alias seems wrong to me. @slurps-mad-rips any comments?

zeux:
Full example that I've tried:

test.cpp:
```
#include "pugixml.hpp"

int main()
{
    pugi::xml_document doc;
}
```

CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.0)
project(test)
add_executable(test test.cpp)
find_package(pugixml)
target_link_libraries(test pugixml)
```

mathstuf:
> I'd rather fix this unconditionally because the intention was to make it possible to depend on pugixml throughout the transition but I'm not sure what the best option to do so is - a circular alias seems wrong to me.

There is no circular alias. the `add_library(pugixml::pugixml ALIAS)` call in `CMakeLists.txt` makes `pugixml::pugixml` work in the pugixml build. The one I'm adding here makes `pugixml` available at `find_package(pugixml)` time (currently, no such target exists).

zeux:
Ahh the change is in the .cmake.in file - I missed that. Would there be any harm in adding this alias unconditionally?

mathstuf:
I think it'd be nice to deprecate the old, un-namespaced, target eventually. The way it is now in this PR is that `find_package(pugixml 1.11)` would *not* provide the old target, but `find_package(pugixml 1.9)` *would* provide it (so that there's only one target name to use).

zeux:
I guess I don't see the potential for the deprecation until 2.0... But this doesn't seem to be harmful so I'll merge this as is. Thanks!

randyfan:
I'm getting the following error: 
image

Here's a screenshot of my CMakeLists.txt which uses find_package(pugixml REQUIRED):
image

Any ideas what's causing the configure error?

mathstuf:
Fixed in #389.

Issue #387, opened 2020-12-08
cmake-ify pugiconfig


hsdk123:
Hi, it would be great to add cmake option variables for each of the defines in `pugiconfig.hpp`.

PR #388, opened 2020-12-09
Cmake-ified pugiconfig

kTindalland:
Attempt at cmake-ifiying the pugiconfig. Wasn't too sure about dllexport, dllimport, and fastcall options.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,44 @@ option(BUILD_SHARED_AND_STATIC_LIBS "Build both shared and static libraries" OFF
  # Technically not needed for this file. This is builtin.
 option(BUILD_SHARED_LIBS "Build shared instead of static library" OFF) 
 
+# Options from the pugiconfig.hpp file.
+
+option(PUGIXML_WCHAR_MODE "Enable wchar_t mode" OFF)
+option(PUGIXML_COMPACT "Enable compact mode" OFF)
+option(PUGIXML_NO_XPATH "Disable XPath" OFF)
+option(PUGIXML_NO_STL "Disable STL" OFF)
+option(PUGIXML_NO_EXCEPTIONS "Disable Exceptions" OFF)
+option(PUGIXML_HEADER_ONLY "Switch to header only" OFF)
+option(PUGIXML_HAS_LONG_LONG "Enable long long support" OFF)
+
+# Tune these to adjust memory-related behaviour
+set(PUGIXML_MEMORY_PAGE_SIZE 32768)
+set(PUGIXML_MEMORY_OUTPUT_STACK 10240)
+set(PUGIXML_MEMORY_XPATH_PAGE_SIZE 4096)
+
+# Tune this to adjust max nesting for XPath queries
+set(PUGIXML_XPATH_DEPTH_LIMIT 1024)
+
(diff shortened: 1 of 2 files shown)


kTindalland:
Intended to resolve #387

zeux:
It would be good to simplify this change:

1) pugiconfig.hpp should not be changed. While it's possible to use CMake as a way to build pugixml, it's nowhere near the only option.
2) Only PUGIXML_WCHAR_MODE and PUGIXML_COMPACT should be exposed. Other options are used for either low-level tuning, various compatibility switches that are best to handle within the CMakeFiles itself (e.g. HAS_LONG_LONG is automatically set), or are irrelevant for CMake completely (HEADER_ONLY). These two are the only ones a user needs to be able to switch pragmatically, and if any others are necessary I'd rather add them later instead of in bulk.

CMakeLists.txt
```diff
@@ -27,6 +27,44 @@ option(BUILD_SHARED_AND_STATIC_LIBS "Build both shared and static libraries" OFF
...
+# Tune this to adjust max nesting for XPath queries
+set(PUGIXML_XPATH_DEPTH_LIMIT 1024)
+
+# Set these to control attributes for public classes/functions (Not 100% sure if these will work.)
+
+option(PUGIXML_OVERRIDE_API "Export all public symbols from DLL" OFF)
+option(PUGIXML_OVERRIDE_CLASS "Import all classes from DLL" OFF)
+option(PUGIXML_OVERRIDE_FUNCTION "Set calling conventions to all public functions to fastcall" OFF)
+
+if(${PUGIXML_OVERRIDE_API})
```
mathstuf, line 54: There's no need to dereference.

```suggestion
if(PUGIXML_OVERRIDE_API)
```

src/pugiconfig.hpp
```diff
@@ -14,41 +14,13 @@
 #ifndef HEADER_PUGICONFIG_HPP
 #define HEADER_PUGICONFIG_HPP
 
-// Uncomment this to enable wchar_t mode
-// #define PUGIXML_WCHAR_MODE
```
mathstuf, line 18: These should probably become `#cmakedefine` calls, the file renamed to `pugiconfig.hpp.in` and then `configure_file`'d in the CMake code.

Drise13:
He all, looking for this functionality. What would it take to get this merged?

zeux:
Addressing comments from https://github.com/zeux/pugixml/pull/388#issuecomment-742613754

src/pugiconfig.hpp
bjornblissing: This should probably be left in place, since some people are building pugixml without using CMake.

So just replacing it with a `#cmakedefine` solution would probably not sufficient.

mathstuf: The current instructions leave non-CMake users to manually edit the file. Would an updated comment for those instructions be sufficient?

bjornblissing: @mathstuf Well, this pull request actually delete all the options from the `pugiconfig.hpp`. 

To make matters worse, the change to the `CMakeLists.txt` do not propagate the option settings to the generated build. 

However, I have created a new pull request which leaves the  `pugiconfig.hpp` unchanged and applies the settings from the `CMakeLists.txt`. See #426

Issue #374, opened 2020-10-21
Suggestion: Prefix pugixml specific CMake options to avoid name collision with other libraries


bjornblissing:
When including pugixml as a submodule to our own project the CMake options and cached variables may result in name collisions.

https://github.com/zeux/pugixml/blob/c4bae8d8185858435bdf22cd98ba7e58e67d9041/CMakeLists.txt#L18-L20

https://github.com/zeux/pugixml/blob/c4bae8d8185858435bdf22cd98ba7e58e67d9041/CMakeLists.txt#L26-L30

My suggestion is to rename the following options and cached variables:
```
BUILD_SHARED_AND_STATIC_LIBS 
BUILD_DEFINES
STATIC_CRT
```

to
```
PUGIXML_BUILD_SHARED_AND_STATIC_LIBS 
PUGIXML_BUILD_DEFINES
PUGIXML_STATIC_CRT
```

The variable `BUILD_SHARED_LIBS` is a CMake built-in variable and probably should be removed. Alternatively defined as:
```cmake
cmake_dependent_option(PUGIXML_BUILD_SHARED_LIBS 
  "Build shared instead of static library" OFF
  "BUILD_SHARED_LIBS" ON)
```

The options below are not created if pugixml is built from outside the pugixml source directory (i.e. another project), so there is no pressing need to prefix those. 
```
USE_VERSIONED_LIBDIR
USE_POSTFIX
BUILD_TESTS
```

hsdk123:
This would be great

zeux:
Fixed by #399.

PR #399, opened 2021-01-11
Add PUGIXML as prefix to CMake options

bjornblissing:
Added PUGIXML as prefix to all CMake options to avoid naming collisions
with downstream projects.

Removed the cached variable BUILD_DEFINES, since it was unused.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,9 @@ cmake_dependent_option(PUGIXML_BUILD_TESTS
   "Build pugixml tests" OFF
   "BUILD_TESTING;CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR" OFF)
 
+# Custom build defines
+set(PUGIXML_BUILD_DEFINES CACHE STRING "Build defines")
+
 option(PUGIXML_BUILD_SHARED_AND_STATIC_LIBS "Build both shared and static libraries" OFF)
  
  # Expose option to build PUGIXML as static even when the global BUILD_SHARED_LIBS variable is set
@@ -42,6 +45,12 @@ endif()
 
 if (NOT DEFINED CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 11)
+endif()
+
+if (DEFINED PUGIXML_BUILD_DEFINES)
+  foreach(DEFINE ${PUGIXML_BUILD_DEFINES})
+    add_definitions("-D" ${DEFINE})
+  endforeach()
 endif()
 
 if (PUGIXML_USE_POSTFIX)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,32 +7,33 @@ include(GNUInstallDirs)
 include(CTest)
 
 
-cmake_dependent_option(USE_VERSIONED_LIBDIR
+cmake_dependent_option(PUGIXML_USE_VERSIONED_LIBDIR
   "Use a private subdirectory to install the headers and libraries" OFF
   "CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR" OFF)
 
-cmake_dependent_option(USE_POSTFIX
+cmake_dependent_option(PUGIXML_USE_POSTFIX
   "Use separate postfix for each configuration to make sure you can install multiple build outputs" OFF
   "CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR" OFF)
 
-cmake_dependent_option(STATIC_CRT
+cmake_dependent_option(PUGIXML_STATIC_CRT
   "Use static MSVC RT libraries" OFF
   "MSVC" OFF)
 
-cmake_dependent_option(BUILD_TESTS
+cmake_dependent_option(PUGIXML_BUILD_TESTS
   "Build pugixml tests" OFF
   "BUILD_TESTING;CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR" OFF)
 
-option(BUILD_SHARED_AND_STATIC_LIBS "Build both shared and static libraries" OFF)
- # Technically not needed for this file. This is builtin.
-option(BUILD_SHARED_LIBS "Build shared instead of static library" OFF) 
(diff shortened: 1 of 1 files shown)


bjornblissing:
This is a PR relating to suggestion #374

zeux:
FWIW the fact that BUILD_DEFINES is unused is a bug, see #413

bjornblissing:
@zeux Do you wish that I amend my pull request by including the re-introduction of the handling of generic build defines via the `BUILD_DEFINES` variable?

zeux:
@bjornblissing Yeah that would be great although I'd imagine you will want that prefixed as well.

bjornblissing:
@zeux I have added a prefixed version of the `BUILD_DEFINES` functionality. However, the need for these types of custom defines may not be needed in the future. The PR #388 is probably the preferred way of handling options available in the library.

zeux:
Thanks! I agree that for PUGIXML_WCHAR_MODE it's better to expose dedicated CMake settings, but for some more arcane options a generic build defines is useful to reduce the configuration space.

CMakeLists.txt
```diff
@@ -43,7 +47,13 @@ if (NOT DEFINED CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 11)
 endif()
 
-if (USE_POSTFIX)
+if (DEFINED PUGIXML_BUILD_DEFINES)
+  foreach(DEFINE ${PUGIXML_BUILD_DEFINES})
+    add_definitions("-D" ${DEFINE})
```
zeux, line 52: I believe this should be using `target_compile_definitions` instead

CMakeLists.txt
```diff
@@ -7,32 +7,36 @@ include(GNUInstallDirs)
...
- # Technically not needed for this file. This is builtin.
-option(BUILD_SHARED_LIBS "Build shared instead of static library" OFF) 
+# Custom build defines
+set(PUGIXML_BUILD_DEFINES CACHE STRING "Build defines")
 
-set(BUILD_DEFINES CACHE STRING "Build defines")
+option(PUGIXML_BUILD_SHARED_AND_STATIC_LIBS "Build both shared and static libraries" OFF)
+ 
+ # Expose option to build PUGIXML as static even when the global BUILD_SHARED_LIBS variable is set
+cmake_dependent_option(PUGIXML_BUILD_SHARED_LIBS
```
zeux, line 32: Why not always provide this option via `option`? Or is the idea that BUILD_SHARED_LIBS overrides this one? Maybe it would make sense to default PUGIXML_BUILD_SHARED_LIBS to the value of the global BUILD_SHARED_LIBS, but still let override in either direction?

CMakeLists.txt
bjornblissing: The difference between `add_definitions` and `target_compile_definitions` is that `target_compile_definitions` applies the setting to a specific target, while `add_definitions` applies the definitions to all targets in this folder and any subfolder. 

Since CMake 3.12 there is also `add_compile_definitions` which adds preprocessor definitions to the compiler command line. (In contrast to `add_definitions` which might include both compiler definitions and compiler flags).

I can certainly change into using `target_compile_definitions`, however the question is if the custom options should be applied to all specified targets (`pugixml`, `pugixml-static`, `pugixml-shared` and `check`) or just a subset of them?

CMakeLists.txt
bjornblissing: Yes, this looks strange. I will change it to your suggestion.

bjornblissing: I agree that this looks strange. However, introducing a local flag `PUGIXML_BUILD_SHARED_LIBS` which defaults to whatever `BUILD_SHARED_LIBS` is set to might be confusing, since this allows you to set the global flag to `TRUE` while the local flag is `FALSE`. Any change to the global flag won't propagate to the local flag, unless you clear the CMake cache. Thus making the behavior a bit unpredictable for the users.

My vote is to remove the statement completely. This will force the users to either use the global flag, OR building the library with the `PUGIXML_BUILD_SHARED_AND_STATIC_LIBS` flag set.

bjornblissing: Another option would be to revert to something similar to the previous version, i.e. exposing the global option (but defaulting to the value from the cache):
`option(BUILD_SHARED_LIBS "Build shared instead of static library" BUILD_SHARED_LIBS)`

However, this might be considered bad form by the CMake purists.

CMakeLists.txt
zeux: I think it should be sufficient to add definitions as PUBLIC to pugixml- targets. This will naturally propagate to all targets that use pugixml, and will make it possible to define ABI-breaking macros like WCHAR_MODE safely. The reason why I suggested the change is my understanding is that add_definitions is part of "legacy" CMake feature set, and modern uses recommend target_ functions.

CMakeLists.txt
zeux: Hmm. If we completely remove this, does this mean `BUILD_SHARED_LIBS` will still work externally since it's a built-in?

CMakeLists.txt
bjornblissing: Fixed in fc7928a4b74df73024f9d8a9698c9e5b46e18a8a

CMakeLists.txt
bjornblissing: I reverted my stance on this and re-introduced the exposure of the global option `BUILD_SHARED_LIBS`. However I put the option `PUGIXML_BUILD_SHARED_AND_STATIC_LIBS` as a dependent option to this. This required me to change the build conditionals a bit to prevent confusing option clashes.

So the commit 7b1b7ee16d86acdb5207966401dccbb28b849475 will allow for three different cases:

1. Static libraries (global `BUILD_SHARED_LIBS` set to `OFF` - *The default behavior*)
2. Shared libraries (global `BUILD_SHARED_LIBS` set to `ON`)
3. Both shared and static (global `BUILD_SHARED_LIBS` set to `ON` and `PUGIXML_BUILD_SHARED_AND_STATIC_LIBS` set to `ON`)

zeux:
Thanks!

PR #406, opened 2021-02-21
fix -Wzero-as-null-pointer-constant warnings

ottigeda:
I am interested what you think about this: would you consider merging this pull-request, or is there a reason to use '0' as null-pointer value ? Whatever the answer will be, thanks for your time!


ottigeda:
After some iterations to make it work I found PUGIXML_NULL,
so I used this to remove this warnings.
What do you think about this ?

zeux:
fwiw I haven't had time to look at the PR yet, and probably won't have a chance in the next couple of weeks - sorry about that. I believe there was some discussion around this a few years ago and the conclusion at the time was that this warning is valuable to stay clean of in the header as this affects user code, hence the use of PUGIXML_NULL, but the maintenance burden inside the actual code is a bit too high. But I'm always happy to reevaluate decisions like this.

We could also consider disabling the warning in the pugixml source code body via compiler specific pragmas when available.

zeux:
Superseded by #586 with smaller scope.

PR #418, opened 2021-04-26
remove const from xml_node_iterator ++/--

neheb:
This prevents usage with C++20 ranges since it does not satisfy
std::weakly_incrementable.

Signed-off-by: Rosen Penev <rosenp@gmail.com>


neheb:
wonder if the other operators should get the same treatment.

zeux:
Did you mean iterators? If so yeah, this should also be changed for `xml_attribute_iterator` and `xml_named_node_iterator`.

neheb:
Yes I did. Removed.

zeux:
Thanks!

PR #419, opened 2021-04-26
add empty method

neheb:
Simple and allows to avoid using std::distance.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/src/pugixml.hpp
+++ b/src/pugixml.hpp
@@ -312,6 +312,8 @@ namespace pugi
 		It begin() const { return _begin; }
 		It end() const { return _end; }
 
+		bool empty() const { return _begin == _end; }
+
 	private:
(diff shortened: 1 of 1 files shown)


src/pugixml.hpp
```diff
@@ -312,6 +312,8 @@ namespace pugi
 		It begin() const { return _begin; }
 		It end() const { return _end; }
 
+		bool empty() const { return _begin == _end; };
```
gnammix, line 315: Nitpicking ,extra, probably unwanted, ```;``` at the end of the line
```suggestion
		bool empty() const { return _begin == _end; }
```

neheb: Fixed.

zeux:
Thanks!

PR #426, opened 2021-06-23
Add Cmake options for pugiconfig.hpp options

bjornblissing:
Add Cmake options variables for each of the defines in pugiconfig.hpp.


bjornblissing:
This is based on the work of @kTindalland in the PR #388. However, that PR changed the pugiconfig.hpp file, which might be problematic for users *not* relying on CMake for their build.

Fixes #387

bjornblissing:
Note that this PR makes the file `pugiconfig.hpp` completely redundant in CMake driven builds.

zeux:
Please see this comment https://github.com/zeux/pugixml/pull/388#issuecomment-742613754 which in part applies to this change as well. I think we should start with a smaller set. wchar mode and compact mode are definitely worth exposing - these are major configuration options. PUGIXML_HAS_LONG_LONG and class/function APIs should be automatically set according to other build settings like shared libraries so these definitely shouldn't be exposed. Various fine grained memory configuration options are too niche to be exposed initially, and we always have PUGIXML_BUILD_DEFINES as a fallback for anything not present here.

bjornblissing:
@zeux One option could use the CMake feature `mark_as_advanced` for the more obscure settings. This would hide these options from GUI users that haven't explicit enabled the `advanced` option in the GUI.

Regarding the fine grained memory configuration options:  These options are not exposed in the GUI at all (since the CMake GUI do not support integer values). The only thing that this PR enables is the ability to set these values via the CMake command line (ie. allows changing these values without editing the `pugiconfig.hpp`.

bjornblissing:
I modified and marked this PR as draft. Currently the following things are solved within this PR:

- `PUGIXML_WCHAR_MODE` and `PUGIXML_COMPACT` are available as standard options in the CMake GUI.
- `PUGIXML_NO_XPATH`, `PUGIXML_NO_STL`, and `PUGIXML_NO_EXCEPTIONS` are available as advanced options in the CMake GUI.
- `PUGIXML_MEMORY_PAGE_SIZE`, `PUGIXML_MEMORY_OUTPUT_STACK`, `PUGIXML_MEMORY_XPATH_PAGE_SIZE`, and `PUGIXML_XPATH_DEPTH_LIMIT` can be set via CMake command line or via parent CMake scripts.
- `PUGIXML_HAS_LONG_LONG` is handled via the CMake `check_type_size` function. *However, this is a bit of duplicate checking since `pugixml.cpp` already to some checks for the availability of `long long` type as well.*

Currently unhandled by this PR:
- There is currently no handling of the option to enable `__fastcall`. I am a bit unsure if this should be handled as an option or as a default on available platforms. 
- There is no way to define `PUGIXML_CLASS` as `__declspec(dllimport)`, since this should only be used by anyone who includes a shared version of the pugixml library. This could be solved by generating a custom `pugiconfig.hpp` header file in the `CMAKE_CURRENT_BINARY_DIR`. However, this may conflict with the option to build pugixml as both static and shared simultaneously.

zeux:
Ok - advanced options for no-stl et al sound good. I don’t think we need to have anything special for custom calling conventions - dll attributes should be setup automatically, x64 only has a single convention, and for arcane uses there’s the define override.

I still think we should remove long long and memory options. Long long override is only there for unusual platforms and deducing that in CMake increases configuration time, and memory options really should be rare to have to tune so I would rather add them if specifically requested in the future. In general it’s easy to add extra options in the future but difficult to remove so would rather start with a small set.

PR #431, opened 2021-06-30
Fix include in test_document.cpp when building against libc++

justinkb:
The _GLIBCXX_HAVE_UNISTD_H relies on pugixml and its tests being built against libstdc++, with LLVM's libc++ this won't be defined. On modern compilers/preprocessors, luckily we can test for includes.

We do so here to include the needed header for unlink, when a) not building for mingw b) we have the __has_include preprocessor facility available and c) the header is in fact available

### DIFF
--- a/tests/test_document.cpp
+++ b/tests/test_document.cpp
@@ -22,12 +22,11 @@
 #	include <stdexcept>
 #endif
 
-#ifdef __MINGW32__
-#	include <io.h> // for unlink in C++0x mode
-#endif
-
-#if defined(__CELLOS_LV2__) || defined(ANDROID) || defined(_GLIBCXX_HAVE_UNISTD_H) || defined(__APPLE__)
-#	include <unistd.h> // for unlink
+// for unlink
+#ifdef _WIN32
+#	include <io.h>
+#else
+#	include <unistd.h>
 #endif
 
 using namespace pugi;


tests/test_document.cpp
```diff
@@ -26,7 +26,7 @@
 #	include <io.h> // for unlink in C++0x mode
 #endif
 
-#if defined(__CELLOS_LV2__) || defined(ANDROID) || defined(_GLIBCXX_HAVE_UNISTD_H) || defined(__APPLE__)
+#if defined(__CELLOS_LV2__) || defined(ANDROID) || defined(_GLIBCXX_HAVE_UNISTD_H) || defined(__APPLE__) || (!defined(__MINGW32__) && defined(__has_include) && __has_include(<unistd.h>))
```
zeux, line 29: I'm wondering, can this simply say `#ifndef _WIN32`? That seems to be effectively what this needs to do since on every non-Windows system, unlink is (probably) in unistd, and the test code doesn't compile without unlink anyhow.

justinkb: most likely, but I used __MINGW32__ because that define is what is used a few lines above this to include io.h.

I checked some mingw headers and their unistd.h is just a stub that includes io.h, so I wanted to avoid including that twice.

justinkb: found some info re: the VS 2015 failure:

https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html

"As a result, combining the two tests into a single expression as shown below would only be valid with a compiler that supports the operator but not with others that don’t". I'll rethink the patch, unless you wanna simplify the whole thing anyway

zeux: Yeah this is why I mentioned that probably replacing this entire condition with #ifndef _WIN32 should be sufficient

justinkb: If that's your preference, I'll implement that. I was wary of making assumptions about what platforms you are supporting with the project, wasn't sure if I could assume unistd.h would be available on every non-windows platform targeted :)

zeux: Yeah I think this is fine here because if unlink doesn't come from unistd.h or io.h it's not clear which header would include it :) If we need to support future non-Windows platforms without unistd, that can be implemented separately.

justinkb: Sure, I guess if someone on some obscure or future platform that doesn't have unistd.h wants to run that test, he'll just have to implement that fix. I was extra wary because I ran into an issue earlier today with some ancient header that got removed from the linux headers in 5.13 (cyclades.h ICYWW). I force-pushed an updated version right before this conversation btw, which seems to be all green. I'll force push the simpler version after the build bot finishes and that'll hopefully be the final version.

justinkb: Alright, testing now with the simple version. I've kept the commit message the same to preserve for the future what prompted the change

zeux:
Thanks!

justinkb:
Cheers!

PR #441, opened 2021-09-29
[WIP] string_view support

halx99:
## Don't needs merge, just a idea
- [x] All string setter function use string_view_t
- [x] API compatible with 1.11.x, but binary incompatible.
- [x] Add explicit boolean type for store bool value as string value `true`, `false` and avoid compiler ambiguous match const char_t* as scalar type 'bool'
- [ ] Add length field to support getter string as string_view_t, it's useful for avoid unnecessary strlen operations, could track this branch: https://github.com/halx99/pugixml/tree/dev

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -5694,7 +5694,7 @@ namespace pugi
 		return impl::strcpy_insitu(_root->value, _root->header, impl::xml_memory_page_value_allocated_mask, rhs.data(), rhs.length());
 	}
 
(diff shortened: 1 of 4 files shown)


brandl-muc:
I think you would need to use `basic_string_view<char_t>` instead of `string_view`.

halx99:
Yes, will change later

halx99:
Close, replaced by the super newer: #457

PR #443, opened 2021-10-01
add meson build support

eli-schwartz:
https://mesonbuild.com/

This should be feature-comparable with the cmake build (it even generates *-config.cmake files for installation, if cmake is installed to acquire the template). It supports all the major OSes, including macOS xcode and Windows visual studio support.

meson also comes with some unique advantages. e.g. its builtin pkg-config generator module allows building third-party projects against an uninstalled developer build using pkg-config *-uninstalled.pc support.

Also, and mainly, projects which use meson can make use of meson's subproject mechanism which allows installing a foo.wrap file into the project, after which dependency lookups for "foo" can retrieve it as an external project and build it in a single configure/compile process. This is very handy for building standalone executables, and somewhat invaluable on platforms that don't have a usable package manager.

Documentation: https://mesonbuild.com/Wrap-dependency-system-manual.html

In order for projects depending on pugixml to make use of this, pugixml needs to be buildable by meson, even if it isn't the default build system.


Neumann-A:
> This should be feature-comparable with the cmake build (it even generates *-config.cmake files for installation, if cmake is installed to acquire the template). It supports all the major OSes, including macOS xcode and Windows visual studio support.

That statement is wrong. The `*-config.cmake` won't work correctly on windows. see https://github.com/mesonbuild/meson/issues/7632

zeux:
I'm going to close this since I don't think I want to maintain one more set of build files. I do not know much about Meson but my understanding is that wrapdb port adequately serves the use case of integrating pugixml into other projects that use Meson. It's also a single source file so hopefully it's also easy to integrate into any given build system like bazel or buck.

PR #442, opened 2021-09-30
[WIP] Improve performance string literal value assignment with shallow copy support

halx99:
## No needs to merge, just a idea
- [x] API compatible with 1.11.x, but binary compatible was break.
- [x] based on branch `halx99/v2-dev`
- [x] Relative PR: https://github.com/halx99/pugixml/pull/1

PR #457, opened 2021-12-23
[WIP] The implements of some v2 concepts for discuss only

halx99:
## CHANGES
- [x] Add length field and update length field at parse & setter time
- [x] All string getter/setter function use string_view_t
- [x] Add explicit boolean type for store bool value as string value `true`, `false` and avoid compiler ambiguous match const char_t* as scalar type 'bool'
- [x] Improve performance string literal value assignment with shallow copy support
- [x] Since getter function return string_view, if you needs get char_t* as 1.11.x, just invoke `string_view::data()`, it never return `nullptr`, this behavior same with v1.11.x

## TODO:
- [ ] Since setter is string_view, some tests can modify to use string_view literal `sv`avoid `strlen`

## sub pull-requests

- #441 
- #442


zeux:
Closing because this should be mostly obsolete as string_view support has been released.

PR #468, opened 2022-01-21
Add support for Visual Studio 2022 with Platform Toolset v143.

appls:
Hi Arseny,
I would like to use pugixml within a project developed in VS 2022 with platform toolset v143. Please have a look at my pull request and if it is working as expected I am happy when you accept it and update the official Nuget package sources.
Cheers Tobias

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,5 @@
 image:
+  - Visual Studio 2022
   - Visual Studio 2019
   - Visual Studio 2017
   - Visual Studio 2015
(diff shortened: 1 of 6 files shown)


KrisWragg:
Would be nice to get an updated nuget package with this in as its currently blocking us updating some projects to VS2022 / VC143 unless we switch to headers only.

Looks like the VS2015 build failure was possibly an intermittent build issue?

```
Error uploading artifact to the storage: The underlying connection was closed: Could not establish trust relationship for the SSL/TLS secure channel.
```

appls:
Hi Kris,

I totally agree with you. My commits should not have any impact on the VS2015 build.
I am also looking forward to get an official package update, since we cannot update our productive systems where the build chain relies on the Nuget package system.
To me it seems that the maintainer, Arseny, is currently busy with other topics and we just have to stay patient.

@zeux Please let me know you need any further support.

Cheers
Tobias

zeux:
I wasn't paying attention to this since the build looked as if it was broken, but after all the fixes it does look like VS2015 failure was spurious.

zeux:
Thanks! I'll look into closing a couple other issues and making 1.12 release with updated NuGet packages

PR #490, opened 2022-04-13
Added bool set_value(const char_t* rhs, size_t sz).

vng:
Useful when creating values from std::string_view


Ferenc-:
> The change itself makes sense but it would be nice to apply the same change to xml_node::set_value as well as xml_text::set.

Hi @zeux,
I have attempted to implement those changes, that you have requested in #522.
Could you take a look?

PureTryOut:
Thanks for merging this @zeux! Is there any ETA for when this gets in a new release?

zeux:
Unsure, likely later this year.

PR #497, opened 2022-05-16
Fix memory leak

### LINKED ISSUE #496: Memory leak in case of status_out_of_memory
Hi
I was trying to load huge xml file and found out that when loading fails with status_out_of_memory some memory wasn't freed. This happens because contents allocated in load_file_impl are not deallocated when impl::convert_buffer returns status_out_of_memory.

TodorHryn:
Fixes #496


zeux:
Thanks! This is a great find.

PR #506, opened 2022-08-04
Enable usage of nullptr for MSVC 16 and newer (MSVS 2010)

brandl-muc:
Support for `nullptr` was added with MS Visual Studio 2010 which has `_MSC_VER == 1600`.

[This Microsoft document](https://docs.microsoft.com/en-us/cpp/porting/visual-cpp-what-s-new-2003-through-2015?view=msvc-170#whats-new-for-c-in-visual-studio-2010) states that `nullptr` support was in fact introduced with 2010, the mapping to `_MSC_VER` was taken from [here](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170).

### DIFF
--- a/src/pugixml.hpp
+++ b/src/pugixml.hpp
@@ -115,6 +115,8 @@
 #ifndef PUGIXML_NULL
 #	if __cplusplus >= 201103
 #		define PUGIXML_NULL nullptr
+#	elif defined(_MSC_VER) && _MSC_VER >= 1600
+#		define PUGIXML_NULL nullptr
 #	else
 #		define PUGIXML_NULL 0
 #	endif


brandl-muc:
We are upgrading to pugixml 1.12.1 and I was surprised when I found out that this is missing.
Sadly branching on `_MSC_VER` is still necessary, on the one hand because older compilers supporting `nullptr` will continue to report `__cpluplus == 199711L` and alas, newer ones by default [still do so](https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/).

zeux:
This check was missing because PUGIXML_NULL was introduced for the benefit of compilers that warn on use of 0 as a null pointer, which MSVC thankfully doesn't. That said, this change seems perfectly good to have as well.

PR #522, opened 2022-09-29
Followup on pr 490

Ferenc-:
Fixes comments in https://github.com/zeux/pugixml/pull/490

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -5691,6 +5691,16 @@ namespace pugi
 		return impl::strcpy_insitu(_root->name, _root->header, impl::xml_memory_page_name_allocated_mask, rhs, impl::strlength(rhs));
 	}
 
+	PUGI__FN bool xml_node::set_value(const char_t* rhs, size_t sz)
+	{
+		xml_node_type type_ = _root ? PUGI__NODETYPE(_root) : node_null;
+
+		if (type_ != node_pcdata && type_ != node_cdata && type_ != node_comment && type_ != node_pi && type_ != node_doctype)
+			return false;
+
+		return impl::strcpy_insitu(_root->value, _root->header, impl::xml_memory_page_value_allocated_mask, rhs, sz);
+	}
+
 	PUGI__FN bool xml_node::set_value(const char_t* rhs)
 	{
 		xml_node_type type_ = _root ? PUGI__NODETYPE(_root) : node_null;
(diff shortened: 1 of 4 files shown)


src/pugixml.cpp
```diff
@@ -5329,11 +5329,16 @@ namespace pugi
...
 	{
 		if (!_attr) return false;
 
-		return impl::strcpy_insitu(_attr->value, _attr->header, impl::xml_memory_page_value_allocated_mask, rhs, impl::strlength(rhs));
+		return impl::strcpy_insitu(_attr->value, _attr->header, impl::xml_memory_page_value_allocated_mask, rhs, sz);
+	}
+
+	PUGI__FN bool xml_attribute::set_value(const char_t* rhs)
+	{
+		return set_value(rhs, impl::strlength(rhs));
```
zeux, line 5341: Should this have the original code as is? Alternatively, maybe other set_value methods should be calling the new overload with size? The asymmetry is odd.

zeux:
Would you mind also adding tests for this to test_dom_traverse.cpp (for xml_node / xml_attribute) and test_dom_text.cpp (for xml_text)?

src/pugixml.cpp
Ferenc-: You are absolutely right.This was never intended... Now I have changed it.

Ferenc-:
> Would you mind also adding tests for this to test_dom_traverse.cpp (for xml_node / xml_attribute) and test_dom_text.cpp (for xml_text)?

I have addedd one to `test_dom_text.cpp` for the xml_text. If you find that TC acceptebale, then I will continue with  `test_dom_traverse.cpp`.

zeux:
> If you find that TC acceptebale, then I will continue with test_dom_traverse.cpp.

Yup something like this is enough, just want to make sure this code is exercised at all by tests.

Ferenc-:
> Yup something like this is enough, just want to make sure this code is exercised at all by tests.

Thanks for the quick feedback!  
I have added one more to `test_dom_text.cpp` and the rest into `test_dom_modify.cpp`  
since I figured that they fit there better, as these are modify functions.

src/pugixml.cpp
zeux: Are you sure this was addressed?

zeux:
> and the rest into test_dom_modify.cpp

Yes, thank you, `test_dom_modify.cpp` is indeed a better place.

Ferenc-:
> > and the rest into test_dom_modify.cpp
> 
> Yes, thank you, `test_dom_modify.cpp` is indeed a better place.

Thanks for your feedback.
Is there anything else I can do to move this forward?

Ferenc-:
Thank you so much for your effort in reviewing, extending and including this.  
Besides cheering me up personally, I believe it opens up new possibilities,
for example the packaging of one particular mapping software into various Linux distros.

PR #525, opened 2022-10-20
Fix exported symbols under clang-cl (Closes #503)

### LINKED ISSUE #503: No symbols exported with clang-cl
Due to:
https://github.com/zeux/pugixml/blob/3b5c1fb022dc286b0480ccafa9b79f832deb24ce/CMakeLists.txt#L114
No symbols get exported if pugixml gets build with clang-cl instead of cl.

bruxisma:
This also turns the define for PUGIXML_API into an `$<IF:>`, instead of an
`$<IF:>` with an empty true condition. If this is inadequate, I will
undo it, and place them on separate lines as they were before, but will
most likely use an inverse `$<NOT:>` instead of an `$<IF:>`.


zeux:
Thanks!

Issue #378, opened 2020-10-29
PugiXML incorrectly handles char8_t in C++20


ytimenkov:
In C++20 there is a `std::u8string` and `char8_t` to make UTF-8 type-safe.

Unfortunately when calling `xml_attribute::set_value(u8"a string")` for example compiler chooses `set_value(bool)` overload which lead to a surprise when all attrubutes became juse `"true"`.

Which makes me thing that providing an overload which accepts `bool` is too wide: it can be any pointer at least. Would be nice to constrain it somehow.

I'm not sure if there is a need to have a separate handling when `pugi::char_t` is `wchar_t` or simply provide such overload when it's a `char`.

ytimenkov:
I also thought that it may be a good idea to just use `char8_t` for `pugixml::char_t` since UTF-8 is used internally anyways.

I think if consumer could define `PUGIXML_TEXT` and `PUGIXML_CHAR` directly instead of relying on `PUGIXML_WCHAR_MODE` things will just work (or provide more knobs...)

zeux:
> Which makes me thing that providing an overload which accepts bool is too wide: it can be any pointer at least. Would be nice to constrain it somehow.

This is sensible; this can probably be achieved using a private overload with `const void*` argument, as you can't use enable_if or other constructs like this due to compatibility requirements. Separately, char8_t as char_t won't work because of reliance on some CRT functions like strcmp.

ytimenkov:
> Separately, char8_t as char_t won't work because of reliance on some CRT functions like strcmp.

Oh, I didn't look that deep, but it feels like case for `std::char_traits<char_t>::compare`...

I guess I should continue sticking to `reinterpret_cast` for now, just use it carefully :)

zeux:
Just noting that I tried to patch this in #719 which by itself is fine (it works) but I realized the problem is more general :( Even before C++20, in PUGIXML_WCHAR_MODE builds it's easy to accidentally call `operator=`/etc. on `const char*`. In default mode, when using pugixml in Windows Unicode applications, I could imagine forgetting to convert from `const wchar_t*`.

All of these could be solved by prohibiting `operator=`/etc. from `const void*`. This could break some existing valid code. However, even that is not enough, because this interacts with unspecified-boolean conversions, which means that you can assign nodes to attributes:

```c++
pugi::xml_node node = doc.append_child("node");
pugi::xml_attribute attr = node.append_attribute("attr");
attr = node;
```

[...]

PR #529, opened 2022-11-02
Add char8_t mode

gix:
### Overview

This PR adds a new mode to pugixml which uses C++20 `char8_t` instead of `char` for the UTF-8 interface. I've dubbed it char8 mode and it can be enabled using the `PUGIXML_CHAR8_MODE` option/macro.

### Motivation

Using pugixml in a mixed ANSI/Unicode Windows codebase is quite errorprone. `char` is already used for the system codepage and it can easily happen that a UTF-8-as-char string slips through without conversion. I've found `char8_t` it to be a huge help in migrating stuff to UTF-8.

### Implementation

Not much has to be changed. Since pugixml already calls C string functions with UTF-8, we can do the same with `char8_t` by just casting in a few places.

Stream-based methods received an additional overload, since the `char` overload may be used to represent arbitrary bytes, and the `char8_t` overload may be used by string streams.

An additional typedef `u8char_t`, which represents the type pugixml uses for a UTF-8 code unit, was added for the conversion functions.

Most changes had to be done in the test code. Representing raw bytes as string literals does not work for UTF-8 literals, since hex escape codes are interpreted as a Unicode character (and replaced by multiple bytes). Affected places either received a branch with a u8 literal or use a new `RAW()` macro which smuggles in UTF-8 code points using chars.

### Notes

- As implemented, the `PUGIXML_CHAR8_MODE` modifies the UTF-8 interface, which means `PUGIXML_WCHAR_MODE` takes precedence if both are defined (no diagnostic).
- I haven't updated the CI build descriptions since I don't know the compiler versions. Locally I've tested with Visual Studio 2022 17.3 and clang 15.
- Maybe an overload taking `char8_t` file paths should be added as well?

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -217,6 +217,8 @@ PUGI__NS_BEGIN
 
 	#ifdef PUGIXML_WCHAR_MODE
 		return wcslen(s);
+	#elif defined(PUGIXML_CHAR8_MODE)
+		return strlen(reinterpret_cast<const char*>(s));
 	#else
 		return strlen(s);
 	#endif
@@ -229,6 +231,8 @@ PUGI__NS_BEGIN
 
 	#ifdef PUGIXML_WCHAR_MODE
 		return wcscmp(src, dst) == 0;
+	#elif defined(PUGIXML_CHAR8_MODE)
+		return strcmp(reinterpret_cast<const char*>(src), reinterpret_cast<const char*>(dst)) == 0;
 	#else
 		return strcmp(src, dst) == 0;
 	#endif
@@ -2300,7 +2304,7 @@ PUGI__NS_BEGIN
 		return wchar_decoder::process(str, length, 0, utf8_counter());
 	}
 
-	PUGI__FN void as_utf8_end(char* buffer, size_t size, const wchar_t* str, size_t length)
+	PUGI__FN void as_utf8_end(u8char_t* buffer, size_t size, const wchar_t* str, size_t length)
 	{
 		// convert to utf8
 		uint8_t* begin = reinterpret_cast<uint8_t*>(buffer);
@@ -2312,13 +2316,13 @@ PUGI__NS_BEGIN
 	}
 
 #ifndef PUGIXML_NO_STL
-	PUGI__FN std::string as_utf8_impl(const wchar_t* str, size_t length)
+	PUGI__FN std::basic_string<u8char_t> as_utf8_impl(const wchar_t* str, size_t length)
 	{
 		// first pass: get length in utf8 characters
 		size_t size = as_utf8_begin(str, length);
 
 		// allocate resulting string
-		std::string result;
+		std::basic_string<u8char_t> result;
 		result.resize(size);
 
 		// second pass: convert to utf8
@@ -3503,7 +3507,7 @@ PUGI__NS_BEGIN
 	#else
 		static char_t* parse_skip_bom(char_t* s)
 		{
-			return (s[0] == '\xef' && s[1] == '\xbb' && s[2] == '\xbf') ? s + 3 : s;
+			return (s[0] == char_t('\xef') && s[1] == char_t('\xbb') && s[2] == char_t('\xbf')) ? s + 3 : s;
 		}
 	#endif
 
@@ -4607,6 +4611,8 @@ PUGI__NS_BEGIN
 	{
 	#ifdef PUGIXML_WCHAR_MODE
 		return wcstod(value, 0);
+	#elif defined(PUGIXML_CHAR8_MODE)
+		return strtod(reinterpret_cast<const char*>(value), 0);
 	#else
 		return strtod(value, 0);
 	#endif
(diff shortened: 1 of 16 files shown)


brandl-muc:
Just a question on the last note. (The rest sound good to me, thanks)
Are you really using or planning to use char8_t for paths? Especially on Windows where UTF-16 is used? There are no overloads for `fopen()` or `std::fstream::open()` AFAIK. Would not `std::filesystem::path` be the better type?

src/pugixml.cpp
```diff
@@ -217,6 +217,8 @@ PUGI__NS_BEGIN
 
 	#ifdef PUGIXML_WCHAR_MODE
 		return wcslen(s);
+	#elif defined(PUGIXML_CHAR8_MODE)
+		return strlen(reinterpret_cast<const char*>(s));
```
Tradias, line 221: Why is strlength not implemented in terms of `std::char_traits<pugi::char_t>::length`, like you did for the tests?

src/pugixml.cpp
```diff
@@ -229,6 +231,8 @@ PUGI__NS_BEGIN
 
 	#ifdef PUGIXML_WCHAR_MODE
 		return wcscmp(src, dst) == 0;
+	#elif defined(PUGIXML_CHAR8_MODE)
+		return strcmp(reinterpret_cast<const char*>(src), reinterpret_cast<const char*>(dst)) == 0;
```
Tradias, line 235: Why is this not implemented in terms of `std::char_traits<pugi::char_t>::compare`?

src/pugixml.cpp
gix: `char_traits` are only available when compiling with STL-support. The tests don't have that restriction.

gix:
Is there any chance of this PR ever progressing?

zeux:
While the changes in the PR itself feel reasonable, I'd like to better understand how widely this will be used. This adds a new compilation mode; for thorough testing it would require increasing number of tested build configurations by 50%. In some sense it's fine, in some other sense it's a notable increase in complexity. It's a C++20 feature so it's not obvious how widely available it would be - I would be more comfortable with some indication from multiple people about this being very valuable in their work.

I think something like this could be merged if there's sufficient demand for a new mode. If there's any way to reduce the differences in tests that would also be nice - this doesn't matter that much for the PR itself, but when adding new tests the wchar_t mode is already a problem because people tend to forget STR macros, so if any of the new macros could be replaced somehow that would be nice. It's not a big deal if they are necessary.

tests/test.hpp
```diff
@@ -154,6 +156,39 @@ struct dummy_fixture {};
...
 
+#ifdef PUGIXML_CHAR8_MODE
+#	if defined(__clang__) || defined(__GNUC__)
+#		define ALIASING_BARRIER(ptr) asm volatile("" : : "rm"(ptr) : "memory")
+#	else
+#		define ALIASING_BARRIER(ptr)
+#	endif
+inline const char8_t* char_cast(const char* bytes)
+{
+	ALIASING_BARRIER(bytes);
```
zeux, line 167: Is this addressing a specific problem? It's really hard to tell without any comments.

zeux:
A little related although that one needs a separate fix to also avoid the footgun in regular mode: https://github.com/zeux/pugixml/issues/378

gix:
> Probably the main thing that would make me comfortable with merging this would be more indication that this is going to be helpful for various users.

For one, it would make pugixml future proof and help adoption of `char8_t`. Many complaints about `char8_t` are about how useless it is since nothing supports it; a classic chicken-and-egg problem. User's on Linux have long misused `char` and treated it as always being UTF-8 but it's not portable (and might not even work on all Linux systems).

On Windows the situation is worse since `char` is almost never UTF-8 and just compiling with `UNICODE` does not make all the legacy APIs go away. The options are:
- `wchar_t`-mode: Increased memory usage. Many unnecessary conversions. UTF-16 is not necessarily a good internal encoding even on Windows.
- UTF-8 as `char`: Error-prone and unportable.
- Convert to `char8_t`: Cumbersome since every use of pugixml looks like `PugiToU8(node.attribute(U8ToPugi(name)).value())`.

GravisZro:
> UTF-16 is not necessarily a good internal encoding even on Windows.

Just thought I would note that the Windows NT kernel and all the Windows core libraries use UTF-16 internally. All other text encodings are converted to from UTF-16. :)

oold:
Regarding UTF-8 vs. UTF-16 on Windows: https://utf8everywhere.org/#windows

It's best for most applications to avoid UTF-16 as much as possible and only use it when interacting with OS APIs. I've sadly experienced way too many code bases where UTF-8, UTF-16, and "system encoding" are all over the place. Most developers have no real understanding of text encodings and should just stick to UTF-8.

P.S.: One such code base uses `wchar_t` everywhere and got ported to Linux, but it only stores UTF-16 in the (4-byte) units in most cases. Why "most cases"? Well, sometimes it's a wide string literal, so it's a real UTF-32 string. Another cautionary tale about using any non-UTF-8 strings in code.

PR #530, opened 2022-11-06
Fix Xcode 14 sprintf deprecation warning

zeux:
We use snprintf when stdc is set to C++11, however in C++98 mode we can't use variadic macros, and Xcode 14 complains about the use of sprintf.

It should be safe however to use variadic macros on any remotely recent version of clang on Apple, unless -pedantic is defined which warns against the use of variadic macros in C++98 mode...

This change fixes the problem for the builds that don't specify -pedantic, which is a problem for another day.


zeux:
I guess technically PUGI__SNPRINTF is only ever used with 4 arguments so we could use a 4-arg version for Xcode 14+ in C++98 mode to work around the problems with -pedantic... It's fragile wrt future changes but it's unlikely that we'll expand the set of calls to snprintf.

PR #533, opened 2022-11-18
Add Github CIFuzz action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where pugixml already runs. CIFuzz can help detect catch regressions and fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -14,7 +14,7 @@ jobs:
       uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
       with:
         oss-fuzz-project-name: 'pugixml'
-        fuzz-seconds: 300
+        fuzz-seconds: 30
         dry-run: false
     - name: Upload Crash
       uses: actions/upload-artifact@v3
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,24 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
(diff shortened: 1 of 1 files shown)


zeux:
Looks like the OSS fuzz run takes ~8 minutes, which extends the GHA runs significantly (from ~3 to ~9 minutes total). Right now AppVeyor is still the long poll but it might change in the future to not run on PRs or run fewer builds.

Would it be reasonable to just run the fuzzing for ~30 seconds instead of ~300? It feels like the value here is in making sure that the fuzzing infrastructure works and not in any actual fuzzing.

DavidKorczynski:
I think changing from 300s to 30s is reasonable considering what type of workflow you prefer! It's nice to have the fuzzers run for just a short while so 30s sounds good to me! Changed it

DavidKorczynski:
We can also speed up the OSS-Fuzz build itself, e.g. these two lines:

https://github.com/google/oss-fuzz/blob/9d3c4ade232dd8b12faa7a269bcf4a5d13545169/projects/pugixml/build.sh#L19-L20

Only creating the `src/pugixml.cpp` object file once should give be some minor improvements. Will fix this up.

zeux:
Great, this is much better. Thanks!

PR #537, opened 2022-12-03
Add overloads with size_t type argument

stefanroellin:
* bool xml_attribute::set_name(const char_t* rhs, size_t sz)
* xml_attribute xml_node::append_attribute(const char_t* name_, size_t sz)
* xml_node xml_node::append_child(const char_t* name_, size_t sz)
* xml_attribute xml_node::attribute(const char_t* name_, size_t sz) const
* xml_node xml_node::child(const char_t* name_, size_t sz) const
* bool xml_node::set_name(const char_t* rhs, size_t sz)

### DIFF
--- a/docs/manual.adoc
+++ b/docs/manual.adoc
@@ -1275,6 +1275,7 @@ As discussed before, nodes can have name and value, both of which are strings. D
 [source]
 ----
 bool xml_node::set_name(const char_t* rhs);
+bool xml_node::set_name(const char_t* rhs, size_t sz)
 bool xml_node::set_value(const char_t* rhs);
 bool xml_node::set_value(const char_t* rhs, size_t size);
 ----
@@ -1297,6 +1298,7 @@ All attributes have name and value, both of which are strings (value may be empt
 [source]
 ----
 bool xml_attribute::set_name(const char_t* rhs);
+bool xml_attribute::set_name(const char_t* rhs, size_t sz)
 bool xml_attribute::set_value(const char_t* rhs);
 bool xml_attribute::set_value(const char_t* rhs, size_t size);
 ----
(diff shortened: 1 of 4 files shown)


zeux:
What's the motivation for these? The change that added `size_t` support for value setup felt reasonable because value is rarely a compile-time literal, and it makes it easier to work with string_view-like inputs. Here I'm not sure that the use is going to be as frequent, and it's a substantial increase of the API surface especially if all other functions that deal with names are included.

stefanroellin:
I work a lot with std::string_view. Since the current interface mostly works with zero-terminated strings, it is not possible to use a std::string_view, since it is not necessarily zero-terminated. Adding the above overloads allows me to use std::string_view.
A solution like implemented in pull request #457 would make this patch obsolete and would suit my needs.

zeux:
I would be willing to merge a subset of this that only introduces set_name. The other additions can be replicated externally without allocations - it’s not always ideal, in particular append_attribute would need to be called with “” followed by set_name - but it’s clean and orthogonal.

Let me know if you’d like to rebase the PR with that subset.

zeux:
Oh also ideally these should be added to docs/manual.adoc although I can do this myself; I can also reduce the PR to a mergeable state myself in a few days if that’s ok.

stefanroellin:
I can rebase the branch and add some documentation. Just to be sure: you will accept a pull request with the following two enhancements:

* bool xml_attribute::set_name(const char_t* rhs, size_t sz)
* bool xml_node::set_name(const char_t* rhs, size_t sz)

stefanroellin:
Branch rebased such that only the two methods

* bool xml_attribute::set_name(const char_t* rhs, size_t sz)
* bool xml_node::set_name(const char_t* rhs, size_t sz)

are included. Also added the methods to the documentation.

zeux:
Thanks!

PR #548, opened 2023-02-27
{cmake} Add option to expose includes as SYSTEM headers

asmaloney:
This can allow projects including pugixml to avoid reporting warnings in pugixml code (depending on compiler) by marking them as SYSTEM headers.

https://cmake.org/cmake/help/latest/command/target_include_directories.html


zeux:
Is this motivated by encountering some specific warnings? The general policy for pugixml is to be warning-free on as many compilers/configurations as possible.

zeux:
Superseded by #550.

PR #550, opened 2023-03-04
Fix weak vtable warning regarding xml_writer

asmaloney:
Using Apple clang (clang-1400.0.29.202) with `-Wweak-vtables` would produce the following warning:

> 'xml_writer' has no out-of-line virtual method definitions; its vtable will be emitted in every translation unit [-Wweak-vtables]

Using the "defaulted default constructor" prevents this.

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -5111,6 +5111,10 @@ PUGI_IMPL_NS_END
 
 namespace pugi
 {
+	PUGI_IMPL_FN xml_writer::~xml_writer()
+	{
+	}
+
 	PUGI_IMPL_FN xml_writer_file::xml_writer_file(void* file_): file(file_)
 	{
 	}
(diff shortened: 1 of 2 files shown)


asmaloney:
Oh - I missed that this failed CI. I didn't get any notification.

It looks like the really old MSVC 2015 doesn't fully support C++11?

> c:\projects\pugixml\tests\../src/pugixml.hpp(327) : error C2065: 'default' : undeclared identifier

zeux:
Yeah, pugixml doesn't require C++11. In this case it should be easy to move the empty definition of xml_writer dtor to pugixml.cpp.

asmaloney:
Ah thanks. I meant to do that, but I switched out to a different project and forgot about this. Will fix it.

zeux:
Thanks!

PR #559, opened 2023-03-19
Fix `-Wreserved-macro-identifier`

SoapGentoo:
* https://eel.is/c++draft/lex.name#3.1
  "Each identifier that contains a double underscore __ [...] is reserved to the implementation for any use."

@zeux We run `-Wreserved-macro-identifier` on our Clang builds at work, and I'd like to upstream the fix we have been carrying locally. The change here is purely mechanical

    sed -i 's|PUGI__|PUGI_DETAIL_|g' src/pugixml.cpp

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -82,39 +82,39 @@
 
 // Inlining controls
 #if defined(_MSC_VER) && _MSC_VER >= 1300
-#	define PUGI__NO_INLINE __declspec(noinline)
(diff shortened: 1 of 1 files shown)


zeux:
This sounds good. Any chance you could rename to PUGI_IMPL? That would match the pugi::impl namespacing used internally.

SoapGentoo:
@zeux done

zeux:
Not in the next month.

PR #573, opened 2023-07-19
{cmake} Fix deprecation warning by updating min version to 3.5

asmaloney:
With CMake 3.27, we get a deprecation warning:

> Compatibility with CMake &lt; 3.5 will be removed from a future version of CMake. Update the VERSION argument &lt;min&gt; value or use a ...&lt;max&gt; suffix to tell CMake that the project does not need compatibility with older versions.

Bump min version to 3.5 to remove warning.


zeux:
Thanks!

PR #576, opened 2023-08-20
Implement parse_merge_pcdata

### LINKED ISSUE #546: Comments split Text into multiple Nodes
If you parse the following XML Node:
```xml
<node>
    Lorem ipsum dolor sit amet,
    consetetur sadipscing elitr, <!-- Some Comment -->
    sed diam nonumy eirmod tempor invidunt <!-- ut labore et dolore magna aliquyam erat-->,
    sed diam voluptua.
</node>
```
you will receive a node with 3 children, that contain `Lorem ipsum dolor sit amet,\nconsetetur sadipscing elitr,`, `sed diam nonumy eirmod tempor invidunt` and `,\nsed diam voluptua.`, respectively.

If you enable `parse_comments`, the root will correctly contain 5 nodes, where the 2 comments are between the 3 text nodes.
But if you disable comments, however, the text will still be split into the 3 pieces, whereas one would probably expect them to be parsed into one single node.

I don't know if that is intended behavior, but IMHO it would be better to be parsed into a single node.

~Jonathan

vineethkuttan:
Enhancement for merging PCDATA if no intermediary nodes are parsed between them.

Fixes #546.


src/pugixml.cpp
```diff
@@ -270,6 +270,18 @@ PUGI_IMPL_NS_BEGIN
 		return static_cast<size_t>(end - s);
 	#endif
 	}
+
+	// concat two strings and store it in the first one 
+	PUGI_IMPL_FN void strconcat(char_t* dst,const char_t* src)
```
zeux, line 275:
```suggestion
	PUGI_IMPL_FN void strconcat(char_t* dst, const char_t* src)
```

src/pugixml.cpp
```diff
@@ -3479,11 +3491,21 @@ PUGI_IMPL_NS_BEGIN
...
 						else
 						{
-							PUGI_IMPL_PUSHNODE(node_pcdata); // Append a new node on the tree.
+							xml_node_struct* cursor_last_child = cursor->first_child ? cursor->first_child->prev_sibling_c + 0 : 0;
 
-							cursor->value = s; // Save the offset.
+							if(PUGI_IMPL_OPTSET(parse_merge_pcdata) && cursor_last_child && PUGI_IMPL_NODETYPE(cursor_last_child) == node_pcdata)
+							{
+								strconcat(cursor_last_child->value, s);
+								s = cursor_last_child->value;
```
zeux, line 3499: I believe this will lead to double unescaping in strconv_pcdata below, which, in addition to being suboptimal for performance, will not be correct for something like this:

```
<node>
&amp;lt; <!-- comment --> &amp;gt;
</node>
```

(the above should result in `&lt;  &gt;` after parsing, but will probably result in `<  &gt;`)

zeux:
I pushed a few commits to clean this up and to add tests for a few flag combinations. I also had to disable this functionality in certain cases when append_buffer is used, as it could result in attempting to concatenate the newly parsed string into a string that was not parsed in the same buffer.

The one issue that remains is that the interaction with `parse_ws_pcdata*` flags is a little odd. I'll need to think about whether this needs to be fixed.

zeux:
I guess one other caveat here is that when this flag is set, the worst case performance of the parser is quadratic in input size because of repeated concatenations :-/

A different implementation strategy where if this flag is set, we do a separate pass after parsing the entire document that merges the nodes might be better here.

zeux:
Alright, I've briefly tried the post-parse tree merging but it has a significant overhead even in the base case where no or a few nodes need to be merged. So I've reverted to the merge-as-we-parse approach suggested here originally, but added a pointer that tracks the last merged fragment which makes merging O(1). This is probably reasonably close now but since the modifications to the parser are fairly invasive I'll need to spend a little more time testing the performance impact here.

zeux:
Ok I think this is as good as it's going to get. I had to tweak a somewhat unrelated part to get back a little perf that was lost in this change, so this seems to be now perf neutral with default flags.

Thanks for kicking this off! I'll merge the PR once the tests pass.

PR #581, opened 2023-09-21
Add inline variable support (external linkage for const vars)

ravstrix:


zeux:
Can you share the motivation for this change?

ravstrix:
Because global constants have internal linkage, each .cpp file gets an independent version of the global variable. Using inline variables solves this problem by using external linkage for such variables. This ensures that only one instance of the variable exists in all files in which it is included.

zeux:
Sure - however, these are numeric constants; clang doesn't emit any symbols here, and gcc doesn't emit any symbols with -O1 or higher. This would be different if the address of the variables is taken, but that should be highly unusual. So I am curious if this is motivated by a specific problem, or is more of a general cleanup.

ravstrix:
It's code cleanup.

zeux:
Thanks!

PR #584, opened 2023-10-06
build: only include CTest when building the pugixml tests

richardapeters:
Currently, the CMakeLists.txt unconditionally includes CTest. As a consequence, BUILD_TESTING is defined by CTest, which has unintended consequences for any project that uses pugixml: a project that has a certain configuration in which the tests are disabled suddenly finds that testing is enabled after including pugixml.

By making include(CTest) conditional on PUGIXML_BUILD_TESTS, BUILD_TESTING is not set anymore in other projects.


zeux:
Thanks!

PR #588, opened 2023-10-16
add some support for std::string_view

ottigeda:
From previous discussions I read here, I assume you will not accept that, since it requires a newer c++ standard.
Anyway I would like to share this with you and get your opinion on it.


JWCS:
I just discovered that std::string wasn't taken by default for many functions, and this seems to be the only instigation of adding this. For the issue of std::string_view, instead of forcefully setting a higher standard, the string_view lines can be wrapped in # if 's checking if the C++ version >= C++ 17. This would support higher versions, without removing support for lower versions. Is there any reason else why this feature hasn't moved forward?

zeux:
I'm not sure if you're asking about this PR specifically or the larger feature. If you're asking about this PR, it increases the standard version requirements, has conflicts and build issues. If you're asking about std::string support, I don't think that would make sense either, as all existing methods can be called if you use `.c_str()`, and this doesn't provide benefits outside of minor ergonomics.

However, if you're asking about string_view, I used to have a mostly negative perspective on this but I think it's now shifted; I will create a new issue outlining my current perspective and what PRs I would be happy to see in that world (I don't have time to do any string_view related work myself).

JWCS:
Thanks, I was asking for the larger feature as a whole. I feel like it's worth adding the few extra wrappers/types for string + string_view, in that it reduces friction when using pugixml. As for me, I was sorely confused when I started seeing pages of compile errors saying "no matching function is defined", when I tried to pass "a string" (string or string_view), and I found that this "modern c++" library only accepted c-strings... 😅 

If it's as simple as adding more boilerplate functions that end up calling the c-string versions, for every function that has overloads for all the data types, it shouldn't be too difficult? Or do you forsee another architecture issue?

zeux:
I don't think pugixml is ever advertised as "modern C++" library. I will summarize the rest elsewhere and link this PR.

zeux:
Closing in favor of strategy outlined in #626 (relevant PRs will be linked to that issue)

PR #589, opened 2023-10-16
add overloads with size_t

ottigeda:
What is your opinion on this addition ?
Those methods are convenient for us, as we have to work with string data which is not necessarily null terminated.


zeux:
Closing in favor of strategy outlined in #626 (relevant PRs will be linked to that issue)

PR #594, opened 2023-11-13
Fix `-Wuseless-cast`

SoapGentoo:
@zeux We build with `-Wuseless-cast` internally, and it would be great to get this upstreamed.

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -1574,8 +1574,8 @@ PUGI_IMPL_NS_BEGIN
 
 		static value_type high(value_type result, uint32_t ch)
 		{
-			uint32_t msh = static_cast<uint32_t>(ch - 0x10000) >> 10;
-			uint32_t lsh = static_cast<uint32_t>(ch - 0x10000) & 0x3ff;
+			uint32_t msh = (ch - 0x10000U) >> 10;
+			uint32_t lsh = (ch - 0x10000U) & 0x3ff;
 
 			result[0] = static_cast<uint16_t>(0xD800 + msh);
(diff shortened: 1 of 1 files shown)


src/pugixml.cpp
```diff
@@ -183,6 +186,12 @@ using std::free;
 #	define PUGI_IMPL_FN_NO_INLINE PUGI_IMPL_NO_INLINE
 #endif
 
+#ifdef PUGIXML_WCHAR_MODE
+#	define PUGI_IMPL_REINTERPRET_CAST_CHAR(ptr) reinterpret_cast<char*>(ptr)
```
zeux, line 190: Can uses of this macro be replaced with `(s - end) * sizeof(char_t)` instead?

src/pugixml.cpp
```diff
@@ -78,6 +78,9 @@
 #	pragma option push
 #	pragma warn -8008 // condition is always false
 #	pragma warn -8066 // unreachable code
+#	define PUGI_IMPL_BORLANDC_CONST_CAST(to, from) const_cast<to>(from)
+#else
+#	define PUGI_IMPL_BORLANDC_CONST_CAST(to, from) from
```
zeux, line 83: Missing #undef at the bottom of the file. (I need to separately look at whether we even still need this because this workaround is probably 15 years old now but for now it's fine to have it wrapped in a macro)

SoapGentoo: can I just remove it instead? That would keep the code simpler and avoid a compat macro

zeux: Yeah that would work

SoapGentoo:
@zeux done

zeux:
Thanks!

PR #605, opened 2024-01-07
Add Bazel support

Vertexwahn:
This PR helps users of the Bazel build system to integrate pugixml in their builds.

Supporting the Bazel build systems is of course more effort to maintain but helps to spread pugixml.
In the worst case the Bazel build path does simply not work - in the best case, someone who uses Bazel can benefit from it.

Also, other projects support Bazel side by side to other build systems (e.g. Catch2, gtest, etc.).

I will maintain the pugixml Bazel build at least for a half year.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,13 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+cc_library(
+    name = "pugixml",
+    srcs = [
+        "src/pugiconfig.hpp",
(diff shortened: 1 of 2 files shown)


zeux:
Thanks for the PR! I do not plan to add more alternative build systems: there’s a few alternatives to CMake that people use, and I would prefer to keep it constrained. External projects are free to define their own build wrappers, add pugixml.cpp to some project (it’s just one file) or use header only mode. See also #443.

PR #620, opened 2024-07-05
Work around 'unreferenced function' warnings in NO_STL builds

### LINKED ISSUE #619: Warnings (C5245) when having PUGIXML_NO_STL and PUGIXML_API defined
Hi,

When compiling pugixml and having the PUGIXML_API/PUGIXML_CLASS definitions set as well as PUGIXML_NO_STL, I get a bunch of C5245 warnings:

```
1>pugixml.cpp(262,22): warning C5245: 'pugi::impl::`anonymous-namespace'::strlength_wide': unreferenced function with internal linkage has been removed
1>pugixml.cpp(2311,22): warning C5245: 'pugi::impl::`anonymous-namespace'::as_utf8_begin': unreferenced function with internal linkage has been removed
1>pugixml.cpp(2317,20): warning C5245: 'pugi::impl::`anonymous-namespace'::as_utf8_end': unreferenced function with internal linkage has been removed
```

I've added ```#   pragma warning(disable: 5245) // unreferenced function with internal linkage has been removed``` to ln54 in the cpp to resolve this. Might be good to either ifdef the functions for ```#define PUGIXML_NO_STL``` or add the warning disable pragma.

zeux:
Normally, as_utf8_begin et al are used in STL functions but when STL is disabled, these are only used if the target platform lacks first class support for wchar_t FILE* APIs. With some warning levels we consequently can get warnings about these functions not being referenced; this affects MSVC (when using aggressive /Wall warning setup) and MinGW (when using -Wall which should work on gcc without warnings, hence the fix).

Not defining these in the first place is difficult because of the complexity of the selection logic for open_file_wide so for now just mark these as unused. The strange (void)& syntax is needed for MSVC to not trigger another warning...

The workaround is narrowly scoped to avoid unforeseen compatibility issues.

Fixes #619.

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -5054,6 +5054,14 @@ PUGI_IMPL_NS_BEGIN
 #if defined(PUGI_IMPL_MSVC_CRT_VERSION) || defined(__BORLANDC__) || (defined(__MINGW32__) && (!defined(__STRICT_ANSI__) || defined(__MINGW64_VERSION_MAJOR)))
 	PUGI_IMPL_FN FILE* open_file_wide(const wchar_t* path, const wchar_t* mode)
 	{
+#ifdef PUGIXML_NO_STL
+		// ensure these symbols are consistently referenced to avoid 'unreferenced function' warnings
+		// note that generally these functions are used in STL builds, but PUGIXML_NO_STL leaves the only usage in convert_path_heap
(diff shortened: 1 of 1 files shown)


zeux:
(one other way to work around this is to mark these inline instead of PUGI_IMPL_FN, which we can do if this workaround ends up being problematic)

PR #623, opened 2024-07-15
throw() -> noexcept

r-barnes:
`throw()` was deprecated in C++11 and removed in C++17. `noexcept` is the appropriate fix.


zeux:
Only deprecated in 17 and removed in 20 according to https://en.cppreference.com/w/cpp/language/noexcept_spec. pugixml doesn’t require C++11 so this will need to be guarded appropriately.

zeux:
PUGIXML_NOEXCEPT should be defined as throw() for builds to pass I believe.

PR #627, opened 2024-09-10
Add string view support (only impl, no tests)

brandl-muc:
Productive implementation should be complete (restricted to the functions you mentioned for a first PR) but there are no tests yet. 
So this is intentionally only a draft to get quick review on the implementation. Despite the limited scope there are still many choices and you might want those differently. Examples are:
* Issue a warning if string_view is requested or not?
* Delegate to existing overload or copy the implementation?
* Define `string_view_t` via a type alias since we already know this is supported or use `typedef` to stay consistent with `char_t` etc.?

I will add tests soon and then publish the PR.

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -5442,7 +5442,7 @@ namespace pugi
 	}
 
 #ifdef PUGI_HAS_STRING_VIEW
-	PUGI_IMPL_FN bool xml_attribute::set_name(const string_view_t rhs)
+	PUGI_IMPL_FN bool xml_attribute::set_name(string_view_t rhs)
 	{
 		return set_name(rhs.data(), rhs.size());
 	}
@@ -5463,7 +5463,7 @@ namespace pugi
 	}
 
 #ifdef PUGI_HAS_STRING_VIEW
-	PUGI_IMPL_FN bool xml_attribute::set_value(const string_view_t rhs)
+	PUGI_IMPL_FN bool xml_attribute::set_value(string_view_t rhs)
 	{
 		return set_value(rhs.data(), rhs.size());
 	}
@@ -5863,7 +5863,7 @@ namespace pugi
 	}
 
 #ifdef PUGI_HAS_STRING_VIEW
-	PUGI_IMPL_FN bool xml_node::set_name(const string_view_t rhs)
+	PUGI_IMPL_FN bool xml_node::set_name(string_view_t rhs)
 	{
 		return set_name(rhs.data(), rhd.size());
 	}
@@ -5890,7 +5890,7 @@ namespace pugi
 	}
 
 #ifdef PUGI_HAS_STRING_VIEW
-	PUGI_IMPL_FN bool xml_node::set_value(const string_view_t rhs)
+	PUGI_IMPL_FN bool xml_node::set_value(string_view_t rhs)
 	{
 		return set_value(rhs.data(), rhs.value());
 	}
@@ -6786,7 +6786,7 @@ namespace pugi
 	}
 
 #ifdef PUGI_HAS_STRING_VIEW
-	PUGI_IMPL_FN bool xml_text::set(const string_view_t rhs)
+	PUGI_IMPL_FN bool xml_text::set(string_view_t rhs)
 	{
 		return set(rhs.data(), rhs.size());
 	}
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -5441,6 +5441,13 @@ namespace pugi
 		return impl::strcpy_insitu(_attr->name, _attr->header, impl::xml_memory_page_name_allocated_mask, rhs, size);
 	}
 
+#ifdef PUGI_HAS_STRING_VIEW
+	PUGI_IMPL_FN bool xml_attribute::set_name(const string_view_t rhs)
+	{
+		return set_name(rhs.data(), rhs.size());
+	}
+#endif
+
 	PUGI_IMPL_FN bool xml_attribute::set_value(const char_t* rhs)
 	{
 		if (!_attr) return false;
@@ -5454,6 +5461,13 @@ namespace pugi
 
 		return impl::strcpy_insitu(_attr->value, _attr->header, impl::xml_memory_page_value_allocated_mask, rhs, size);
 	}
+
+#ifdef PUGI_HAS_STRING_VIEW
+	PUGI_IMPL_FN bool xml_attribute::set_value(const string_view_t rhs)
+	{
+		return set_value(rhs.data(), rhs.size());
+	}
+#endif
 
 	PUGI_IMPL_FN bool xml_attribute::set_value(int rhs)
 	{
@@ -5848,6 +5862,13 @@ namespace pugi
 		return impl::strcpy_insitu(_root->name, _root->header, impl::xml_memory_page_name_allocated_mask, rhs, size);
 	}
 
+#ifdef PUGI_HAS_STRING_VIEW
+	PUGI_IMPL_FN bool xml_node::set_name(const string_view_t rhs)
+	{
+		return set_name(rhs.data(), rhd.size());
+	}
+#endif
+
 	PUGI_IMPL_FN bool xml_node::set_value(const char_t* rhs)
 	{
 		xml_node_type type_ = _root ? PUGI_IMPL_NODETYPE(_root) : node_null;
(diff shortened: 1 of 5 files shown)


src/pugixml.cpp
```diff
@@ -5441,6 +5441,13 @@ namespace pugi
 		return impl::strcpy_insitu(_attr->name, _attr->header, impl::xml_memory_page_name_allocated_mask, rhs, size);
 	}
 
+#ifdef PUGI_HAS_STRING_VIEW
+	PUGI_IMPL_FN bool xml_attribute::set_name(const string_view_t rhs)
```
zeux, line 5445: Here and elsewhere `const` should not be used for arguments as they are passed by value

src/pugixml.hpp
```diff
@@ -122,6 +122,26 @@
...
+#		define PUGI_STRING_VIEW_SUPPORTED 1
+#	endif
+#endif
+// Enable 'string_view' support if requested and supported
+#ifndef PUGI_HAS_STRING_VIEW
+#	if defined(PUGI_ENABLE_STRING_VIEW) && !defined(PUGIXML_NO_STL)
+#		ifdef PUGI_STRING_VIEW_SUPPORTED
+#			define PUGI_HAS_STRING_VIEW
+#		else
+#			warning "Support for std::string_view was requested but is not supported by your C++ standard"
```
zeux, line 137: #warning is C++23. It's probably fine to keep this undiagnosed.

src/pugixml.hpp
```diff
@@ -140,6 +160,11 @@ namespace pugi
 	// String type used for operations that work with STL string; depends on PUGIXML_WCHAR_MODE
 	typedef std::basic_string<PUGIXML_CHAR> string_t;
 #endif
+
+#ifdef PUGI_HAS_STRING_VIEW
+	// String view type used for overloads of functions that accept strings; depends on PUGIXML_WCHAR_MODE
+	using string_view_t = std::basic_string_view<char_t>;
```
zeux, line 166: typedef is probably better for consistency

src/pugixml.hpp
```diff
@@ -122,6 +122,26 @@
 #	endif
 #endif
 
+// Detect if C++ standard supports 'string_view' (2017 or higher)
```
zeux, line 125: I would propose simplifying this to enabling internal PUGI_HAS_STRING_VIEW define if `PUGIXML_STRING_VIEW` define (note the `PUGIXML` prefix; `PUGI_` prefix is for internal macros, `PUGIXML_` prefix is for user facing configuration) is enabled, and dropping the `_SUPPORTED` macro as it does not seem to be useful.

Separately, `PUGI_HAS_STRING_VIEW` should be undef'd at the end of the header.

`PUGIXML_STRING_VIEW` should be added, commented out, in `pugiconfig.hpp`. We can note there that the macro will become unnecessary with future versions.

src/pugixml.hpp
```diff
@@ -122,6 +122,26 @@
...
+#ifndef PUGI_STRING_VIEW_SUPPORTED
+#	if defined(__cpp_lib_string_view) || __cplusplus >= 201703L || defined(_MSVC_LANG) && _MSVC_LANG >= 201703L
```
zeux, line 127:
```suggestion
#	if defined(__cpp_lib_string_view) || __cplusplus >= 201703L || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
```

zeux: Also, out of curiosity, what is the distinction between _cpp_lib & __cplusplus checks here? Do we need both and if so, for what purpose?

zeux:
This is a good isolated starting point (modulo tests and adding the define to GitHub Actions / AppVeyor config). I think it's fine to delegate to existing overloads when they are available; these details could be changed later if need be (perhaps for debug performance, although using string_view necessitates function calls in debug builds anyhow).

src/pugixml.hpp
brandl-muc: Oh thanks, that explains a lot, I was wondering why I hadn't discovered it earlier.

src/pugixml.hpp
brandl-muc: A question regarding consistency: `PUGI_HAS_STRING_VIEW` seems to be the only macro using `PUGI` prefix *in the header*. All others use `PUGIXML`, e.g., `PUGIXML_HAS_MOVE` or `PUGIXML_HAS_LONG_LONG` which are auto-detected as well. `PUGI`-macros can only be found in the cpp.
So shouldn't it be `PUGIXML_HAS_STRING_VIEW`?

Probably for the same reason it would be the only macro that is actually undefined in the header. (there is no `#undef` to be found) So I cannot orient on anything. Should I always `#undef` it even if I didn't defined it? (not that there is a `#ifndef PUGI_HAS_STRING_VIEW`)

src/pugixml.hpp
brandl-muc: In principle the [feature testing macros](https://en.cppreference.com/w/cpp/feature_test) are a standardized way of  directly testing feature support. If they are available then it is less hassle than testing `__cplusplus` (thanks to `_MSVC_LANG`) and they are more expressive.
Sadly, they are only guaranteed with C++20 although the big three provide them earlier. (if I researched correctly)

 So `__cpp_lib_string_view` is neither required nor sufficient.

I wrote it like that because [someone on Microsoft's STL implementation](https://github.com/microsoft/STL/issues/2503#issuecomment-1021063613) proposed this.
And I can see one advantage. If your compiler does not support the entirety of C++17 but already supports `string_view` and defines the macro accordingly then support would be enabled.

brandl-muc: So seeing that the options are to either remove the check on `__cpp_lib_string_view` or to also add
```
#if defined(__has_include) && __has_include(<version>)
#   include <version>
#endif
```
gives a pretty good idea which option you'll prefer.

src/pugixml.cpp
brandl-muc: This is my usual style (as much `const` as possible), although I would use east const.
I don't have any problems with changing this, but out of curiosity, could you elaborate on your rationale?
The fact that they are passed by value is orthogonal to the fact that the value is not changed.
And because it is not changed it can be made `const` and IMHO consequently should be made `const`.

zeux: I personally do not subscribe to the "what could be made const should be" when that doesn't affect the interface; in this case, the caller behavior does not change based on whether the argument is const or not. I am aware of the argument that it can help prevent mistakes but I have not seen these happen much in practice so I view this as syntactic noise. You're welcome to disagree of course, but for consistency this `const` should be removed: pugixml doesn't have a single function with a const parameter that is passed by value, and the only uses of const locals are that where the argument is a *compile-time* const (when this code was written this seemed to  enhance readability, whether this is true or not is subjective).

src/pugixml.hpp
zeux: Apologies, I've misled you about the prefix; I forgot the conventions :) Indeed, it should be prefixed with `PUGIXML_` and no undef is necessary. I mixed this up with the `.cpp` file where the prefix is `PUGI_IMPL_` and the macros are undef'd to not pollute the namespace in header-only mode.

brandl-muc:
Regarding "adding the define to GitHub Actions / AppVeyor config", this seems to be the next step, before adding unit tests.
A gave it a shot but you will probably not like it, so please help with that.

zeux:
> For gcc/clang my approach seems to work, for Windows I'm not so sure: (and I can't test locally)

For all of them, this should be an expansion of the existing matrix (just need to add this to defines).
For Makefile builds, you should not need to change the Makefile itself. defines is already providing you what you need. You do indeed need to add a make invocation with c++17 standard, probably `debug` config is better because that will maybe enable more assertions and be faster to build.
For CMake builds, you need to add handling of this define to CMakeLists.txt; unsure if you need to override the standard version for MSVC to work, but that can be done via `CMAKE_CXX_STANDARD` externally if necessary I guess.

brandl-muc:
But wouldn't adding `PUGIXML_STRING_VIEW` to the matrix never build it in combination with say `PUGIXML_WCHAR_MODE`?

zeux:
If you want that combination I think it can be added by adding `"PUGIXML_WCHAR_MODE,PUGIXML_STRING_VIEW"`. In general pugixml doesn't test all permutations on GHA as there are too many; I could see an argument for this specific combination being more valuable than most I guess, although I would also assume that we don't have to special case this in the code necessarily.

brandl-muc:
You're probably right, WCHAR_MODE is not really that different when it comes to the string_view changes.
I'll try to rework the integration soon.

dantargz:
Hi, it looks like this PR has gotten stuck. I took a stab at adapting all of the tests that were added for the char*+size versions of the functions to also test string_view, updating the cmake files, and updating the workflow files to add coverage.

I am new to making open source contributions, I didn't realize that making a draft pr in my fork would add a message to this thread, sorry for the noise. Let me know if I can help in some way, like by merging any testing / cmake / build.yml changes into brandl's fork for inclusion in this PR.

brandl-muc:
Hi @dantargz, you mentioned this PR in the description of your PR, that's what caused "the noise".
I hadn't forgotten this PR but apple harvest (I live on a small former farm) required my spare time. But help of course is appreciated, I would drop mine when your PR gets merged.

dantargz:
thanks, I opened a PR against the main repo here: https://github.com/zeux/pugixml/pull/633

zeux:
Merged as #633, thanks!

Issue #626, opened 2024-09-05
Perspective on std::string_view support


zeux:
I think it would be helpful to outline my current perspective on `std::string_view`; it is different from what it has been. Doing this in a separate issue to make this more accessible, but this was originally requested many years ago in #73; also a few related PRs have been filed over the years, #441, #495, #588.

This was originally requested before `std::string_view` became standard and common-place. At the time, my hope was that the library interface can eventually be reworked to switch from `char*` in the interface towards some sort of wrapper like `pugi::string_arg`. This wrapper could hide the complexity of dispatch and take `std::string`, `std::string_view`, `const char*`, and whatever else when constructed. Without this, support for string_view would be limited to either pre-release (when requested originally) or very recent (when requested again :sweat_smile:) C++ standards; since these are not always universally available, I was concerned that a significant expansion of the interface (every function that takes `const char*` needs to be duplicated!) would be rarely used. For C++ standards that *didn't* have `std::string_view`, this was also unfortunate because `std::string` also exists but having 3 overloads would be even more onerous.

So the mental model at the time was "string_arg will solve this". However, `string_arg` would be an ABI-breaking change (also in some cases an API-breaking change for people who rely on implicit conversions to `const char*`, but hopefully this is rare!), so something that would happen after 1.x. But 1.x lasted for way longer than I thought because I'm averse to major compat breaking changes anyway, and this alone seemingly didn't justify it.

Meanwhile, people started increasingly adopting `std::string_view`, so what used to be a minor annoyance if you had a `std::string` (just call `.c_str()`!) became more of a hurdle. Exposing pugixml via an interface to any language that doesn't use null terminated strings was similarly problematic.

A lot of pugixml methods like `child(const char*)` are mere helpers, and can be reimplemented without the use of any method that takes a string, by iterating through children externally. Some, like `set_name()`, can't be reimplemented so they would require an allocation-and-copy at the call site, which is problematic. To address that, over the last two releases (v1.13 in 2022 and v1.14 in 2023) additional overloads for all setters that take `const char*, size_t` pair were added. This doesn't solve ergonomics, but at least it solves the fundamental inability to use string_view like types... the problem of course is you need to stop using some methods like `child(const char*)`, which reduces the ease of use - which is valuable for pugixml.

[...]

PR #633, opened 2024-10-22
Initial support for std::string_view

dantargz:
Based on work from @brandl-muc 

Initial PR for supporting std::string_view as laid out in https://github.com/zeux/pugixml/issues/626
* Add a define guard to make initial availability opt-in
* Add includes, typedefs, and support for WCHAR mode
* Add overloads for set_name/set_value/set functions, which already have support for const char* + size
* Add changes to build files & CI configurations
* Adapt existing unit tests for set_name/set_value/set overloads that take const char* + size to also exercise string_view path, including exercising cases where the data pointer in the string_view is nullptr

For CI config, the new define was added to the matrix testing configuration. The new code is only exercised when using c++17 std lib or later, so an additional build & run was added to all platforms. After the code settles, the expectation is for PUGIXML_STRING_VIEW to be retired, which will help reduce CI run bloat.

Appveyor configs are not updated as part of this change. The appveyor runs are determined to be slow, the majority of new runs created by adding this to the matrix  would be redundant due to most of the matrix not having c++17, and we will gain coverage automatically when the opt-in define is retired.

This change causes a new internal define `PUGIXML_HAS_STRING_VIEW` to be conditionally defined by the header, and it is never undefined.


.github/workflows/build.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       matrix:
         os: [ubuntu, macos]
         compiler: [g++, clang++]
-        defines: [standard, PUGIXML_WCHAR_MODE, PUGIXML_COMPACT, PUGIXML_NO_EXCEPTIONS]
+        defines: [standard, PUGIXML_WCHAR_MODE, PUGIXML_COMPACT, PUGIXML_NO_EXCEPTIONS, PUGIXML_STRING_VIEW]
```
dantargz, line 15: I am not familiar with appveyor -- please advise what changes you would recommend (if any) for that system.

zeux: New define could be added to tests/autotest-appveyor.ps1 line 40. This by itself might not be enough due to C++17 restriction; Appveyor tests a variety of compiler versions most of which predate C++17 but that's fine as these would be tested via GitHub Actions.

src/pugixml.hpp
```diff
@@ -38,6 +38,20 @@
 #	include <string>
 #endif
 
+// Check if std::string_view is both requested and available
+#if defined(PUGIXML_STRING_VIEW) && !defined(PUGIXML_NO_STL)
+#	if __cplusplus >= 201703L
+#		define PUGI_HAS_STRING_VIEW
```
zeux, line 44: There was some confusion in the original PR wrt naming; this should be `PUGIXML_HAS_STRING_VIEW` as `PUGI_` prefix is not actually used except for defines internal to `pugixml.cpp`

.github/workflows/build.yml
dantargz: sounds good, I'll add a note about this in the description. Once the opt-in define is retired, appveyor builds will start exercising this automatically in the appropriate scenarios.

src/pugixml.hpp
dantargz: Sound good, I will rename. It occurred to me after sending the PR that this check is not too complicated -- if we wanted to avoid leaking the new define, we could copy-paste this check into the .cpp and undefine at the end of the header and .cpp files. Leaving it as is also works, as there is already precedent for the header adding PUGIXML_ defines.

zeux: I think it's fine to keep the define for now since as you mention there's already other similar defines, and some are also gated on C++ version.

zeux:
Thanks!

PR #641, opened 2024-10-27
Update VERSION in Makefile and add documentation for string_view_t

dantargz:
Add information to the docs about `PUGIXML_STRING_VIEW`, `PUGIXML_HAS_STRING_VIEW`, `string_view_t` and the overloads that use it. Attempted to have the documentation be future-leaning by primarily documenting the (permanent) `PUGIXML_HAS_STRING_VIEW` define and only mentioning `PUGIXML_STRING_VIEW` as a note in the documentation for `PUGIXML_HAS_STRING_VIEW`.

Docs generated with newer version of asciidoctor 2.0.23 which seems to have also moved some of the css around, though I have not noticed a visual difference.

The `VERSION` variable in the Makefile was accidentally broken in a previous commit by adding the word "version" to pugixmlconfig.hpp. This is the output of the sed command used to define `VERSION`

The result of the old sed command:
```
% sed -n 's/.*version \(.*\).*/\1/p' src/pugiconfig.hpp 
1.14
of pugixml this macro will become obsolete.
```

which caused this behavior when trying to build docs
```
% make docs/manual.html   
asciidoctor -b html5 -a version=1.14 of pugixml this macro will become obsolete. docs/manual.adoc -o docs/manual.html
asciidoctor: FAILED: input file of is missing
make: *** [docs/manual.html] Error 1
```

Tweaked the sed command to /q quit after the first occurance of the word "version" to handle this case.


docs/manual.adoc
```diff
@@ -244,6 +244,8 @@ NOTE: In that example `PUGIXML_API` is inconsistent between several source files
 
 [[PUGIXML_HAS_LONG_LONG]]`PUGIXML_HAS_LONG_LONG` define enables support for `long long` type in pugixml. This define is automatically enabled if your platform is known to have `long long` support (i.e. has C{plus}{plus}11 support or uses a reasonably modern version of a known compiler); if pugixml does not recognize that your platform supports `long long` but in fact it does, you can enable the define manually.
 
+[[PUGIXML_HAS_STRING_VIEW]]`PUGIXML_HAS_STRING_VIEW` define enables function overloads that accept `std::basic_string_view` arguments. This define is automatically enabled if built targeting c++17 or later AND if `PUGIXML_STRING_VIEW` is also defined. The requirement to additionally define `PUGIXML_STRING_VIEW` will be retired in a future version.
```
zeux, line 247:
```suggestion
[[PUGIXML_HAS_STRING_VIEW]]`PUGIXML_HAS_STRING_VIEW` define enables function overloads that accept `std::basic_string_view` arguments. This define is automatically enabled if built targeting C++17 or later AND if `PUGIXML_STRING_VIEW` is also defined. The requirement to additionally define `PUGIXML_STRING_VIEW` will be retired in a future version.
```

zeux (approved):
Thanks!

PR #642, opened 2024-10-30
Auto-detect std::string_view support by default

### LINKED ISSUE #640: Automatically enable std::string_view support
In this release, we're going to be adding support for `std::string_view` (see #626); to enable it, you'd need:

- to enable `PUGIXML_STRING_VIEW` define
- to make sure pugixml is using C++17 during compilation (and your compiler supports it)

The need to enable the define is intentional to minimize compatibility risk; however, in the next release after that the intent is to enable this by default. To do this, we'll just remove the use of the define, but the language support will need tweaks in CMake which currently sets the language to 11 unconditionally unless overridden externally. We probably need to instead check if the target compiler supports C++17 and enable it then; we could also set 17 and remove the code that enables C++ version requirements, but that may have undesirable side effects.

The language setting in CMake also should not propagate to dependent projects: not all dependent code may compile as C++17.

Makefile builds will already inherit the default C++ version from the compiler after #639, which effectively enables C++17 for latest versions of gcc/clang, but we can't just rely on the defaults in CMake as MSVC still doesn't enable C++17 by default.

zeux:
This PR enables std::string_view by default when the compiler supports C++17, and also tries to enable C++17 in a few cases when the compiler should support it. Notably, vcxproj files for VS2019/2022 now enable /std:c++17, and CMake enables C++17 when no specific version is set via CMAKE_CXX_STANDARD; we are now using CMAKE_CXX_STANDARD_REQUIRED=OFF to use CMake's transparent C++ version downgrade.

I originally expected that the next release should go out with string_view being opt-in. However, I'm thinking it might be better to just go all-in:

1. The release doesn't really have much else outside of a few compatibility fixes. Even if there is a compatibility regression, the only way to find out is to try, and if the release has issues this wouldn't block any other features.
2. During testing, I encountered a few surprises wrt enabling C++17 support, but zero actual compatibility issues; maybe the worst case here is "std::string_view could be supported but isn't", and risk is reduced.

If this does go in, we'd delay the release; I was originally hoping for a November release, but might push this to January 2025 if this goes through to allow for testing and last minute fixes to go in.

Relates to #626.
Closes #640.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         os: [ubuntu, macos]
         compiler: [g++, clang++]
-        defines: [standard, PUGIXML_WCHAR_MODE, PUGIXML_COMPACT, PUGIXML_NO_EXCEPTIONS, PUGIXML_STRING_VIEW]
+        defines: [standard, PUGIXML_WCHAR_MODE, PUGIXML_COMPACT, PUGIXML_NO_EXCEPTIONS]
         exclude:
           - os: macos
             compiler: g++
@@ -39,7 +39,7 @@ jobs:
     strategy:
       matrix:
         arch: [Win32, x64]
-        defines: [standard, PUGIXML_WCHAR_MODE, PUGIXML_COMPACT, PUGIXML_NO_EXCEPTIONS, PUGIXML_STRING_VIEW]
+        defines: [standard, PUGIXML_WCHAR_MODE, PUGIXML_COMPACT, PUGIXML_NO_EXCEPTIONS]
     steps:
     - uses: actions/checkout@v1
     - name: cmake configure
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,11 +64,13 @@ if (NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY)
     MultiThreaded$<$<CONFIG:Debug>:Debug>$<$<NOT:$<BOOL:${PUGIXML_STATIC_CRT}>>:DLL>)
 endif()
 
-if (NOT DEFINED CMAKE_CXX_STANDARD_REQUIRED)
-  set(CMAKE_CXX_STANDARD_REQUIRED ON)
-endif()
+# Set the default C++ standard to C++17 if not set; CMake will automatically downgrade this if the compiler does not support it
+# When CMAKE_CXX_STANDARD_REQUIRED is set, we fall back to C++11 to avoid breaking older compilers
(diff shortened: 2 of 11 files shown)


zeux:
Setting `CMAKE_CXX_STANDARD` is not that great here: the choice propagates to dependent libraries, so this would change behavior for dependent projects that don't specify the standard. I guess ideally we'd enable C++17 for pugixml target without forcing this choice on dependent projects, so they can use `std::string_view` if they have enabled C++17 themselves.

mosra:
(Sorry in advance if this is unhelpful or too tangential.)

What I do in [Magnum](https://github.com/mosra/magnum) regarding opt-in C++14/17/20 support (or, opt-in STL support in my case) is that I let the base library be always compiled with some conservative default, and the opt-in functionality is fully inlined in a header (or even tucked away into an extra header if it'd mean pulling in a lot of extra weight). Then, in order to verify that the opt-in functionality actually works, I have dedicated test exectuables that are compiled with appropriate higher standard version.

That way I don't need to worry about accidental ABI breakages, which theoretically *shouldn't* happen from just switching a C++ standard, but I remember rare cases where attempting to link a library built with C++17 to a C++11 codebase led to issues due to the library using some libstdc++ internals the C++11 build had no idea about. It also makes dealing with packages/installations easier, because even if a distribution builds with an older C++ standard for whatever reason, it doesn't prevent users from actually using the C++17 functionality, if it's all inline.

In the case here it'd mean that the code would inline all operations involving a `std::string_view`, and they'd delegate to the data + size overloads:

```cpp
    bool set_value(const char_t* rhs);
    bool set_value(const char_t* rhs, size_t size);
#ifdef PUGIXML_HAS_STRING_VIEW
    bool set_value(string_view_t rhs) {
        return set_value(rhs.data(), rhs.size());
    }
#endif
```

From my personal experience, given the size of `<string_view>` itself, the extra inlined methods won't really make compile times any worse. Assuming the compiler inlines at least a bit, for this particular case I think the call overhead would be equivalent to when `set_value(string_view_t)` was defined in the `*.cpp`, where it was basically a copy of `set_value(const char_t*, size_t)`. In any case, the major advantage is that you then don't need to worry about enabling C++17 when building the library itself.

zeux:
Yeah - I considered this before but this is a further increase in surface area as not all methods with string_view overloads currently have size_t equivalents (in fact most don't), and it would be simpler to keep it that way.

I did check the ABI consequences and it doesn't look like this would break it on any mainstream platforms (Linux/Windows/macOS). I'm also assuming that the distributions will package without overriding the CXX standard, so since it defaults to 17 it should be fine regardless of whether the user requests C++11 or C++17.

PR #644, opened 2024-10-31
add C++20 module

Arthapz:


brandl-muc:
Given that constexpr has evolved substantially it might be a good idea to name the macro PUGIXML_CONSTEXPR11.
That would allow e.g. to have functions as constexpr in C++17 mode that could not be constexpr with C++11.

Arthapz:
> Given that constexpr has evolved substantially it might be a good idea to name the macro PUGIXML_CONSTEXPR11. That would allow e.g. to have functions as constexpr in C++17 mode that could not be constexpr with C++11.

good catch

Arthapz:
here's a demo for C++20 module

https://github.com/Arthapz/pugixml_module_test

aniol-marin:
What's the state on this PR? Any chance it can be discussed further?

Arthapz:
(i rebased on master)

mark-z-r:
Would like to use this feature... is there a reason it isn't being merged?

aniol-marin:
> Would like to use this feature... is there a reason it isn't being merged?

I assume modules are not a priority right now. Maybe we could maintain the modules fork ourselves in order to keep it tested for then @zeux is ready to tackle module support.

PR #653, opened 2025-01-02
Add support for Apple framework builds

Treata11:
Add CMake option `PUGIXML_BUILD_APPLE_FRAMEWORKS` to build PugiXML as Apple Frameworks.
This is needed to be able to create iOS-derived compatible app bundles which need to embed PugiXML as framework, as using dylibs is prohibited by Apple.

CMake officially made support for framework bundles in v3.14:
https://cmake.org/cmake/help/latest/prop_tgt/FRAMEWORK.html#prop_tgt:FRAMEWORK

**Configure** (iOS):
```bash
cmake -S. -B build -DPUGIXML_BUILD_APPLE_FRAMEWORKS=TRUE -DPUGIXML_INSTALL=TRUE -DCMAKE_SYSTEM_NAME=iOS
```
**Build**:
```bash
sudo cmake --build build --config Release --target install
```

Built frameworks would be written at: `build/pugixml.framework`


CMakeLists.txt
```diff
@@ -50,6 +50,10 @@ option(PUGIXML_NO_STL "Disable STL" OFF)
 option(PUGIXML_NO_EXCEPTIONS "Disable Exceptions" OFF)
 mark_as_advanced(PUGIXML_NO_XPATH PUGIXML_NO_STL PUGIXML_NO_EXCEPTIONS)
 
+if (APPLE)
+  option(PUGIXML_BUILD_APPLE_FRAMEWORKS "Build as Apple Frameworks" OFF)
```
zeux, line 54: Maybe `PUGIXML_BUILD_APPLE_FRAMEWORK` (singular) would make more sense?

CMakeLists.txt
```diff
@@ -234,7 +248,8 @@ if(PUGIXML_INSTALL)
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT ${PUGIXML_RUNTIME_COMPONENT}
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT ${PUGIXML_LIBRARY_COMPONENT} ${namelink-component}
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT ${PUGIXML_DEVELOPMENT_COMPONENT}
-    INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}${versioned-dir})
+    INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}${versioned-dir}
+    FRAMEWORK DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT runtime OPTIONAL)
```
zeux, line 252: What will this do on earlier versions on CMake? I am assuming the option doesn't need to work obviously, but unsure if CMake 3.10 will break with this when install is used because FRAMEWORK is unknown.

CMakeLists.txt
Treata11: Sure!

zeux:
Thanks!

PR #655, opened 2025-01-11
Include Headers in the lib

Treata11:
Greetings, 

When I built the library for iOS, I realized that there were no `Header` directories included in the `framework` bundle.
```bash
cmake -S. -B _build -DPUGIXML_BUILD_APPLE_FRAMEWORK=TRUE -DPUGIXML_INSTALL=TRUE -DCMAKE_SYSTEM_NAME=iOS -DPUGIXML_USE_VERSIONED_LIBDIR=ON

sudo cmake --build _build --config Release --target install
```
I searched the entire build dir & didn't find any trace of the headers.

So, I included the headers in the lib & also added them in the framework's bundle.


CMakeLists.txt
```diff
@@ -182,8 +189,10 @@ if (PUGIXML_BUILD_APPLE_FRAMEWORK)
   set_target_properties(${libs} PROPERTIES
     FRAMEWORK TRUE
     FRAMEWORK_VERSION ${PROJECT_VERSION}
-    XCODE_ATTRIBUTE_PRODUCT_BUNDLE_IDENTIFIER com.zeux.pugixml
-    MACOSX_FRAMEWORK_IDENTIFIER com.zeux.pugixml
+    XCODE_ATTRIBUTE_PRODUCT_BUNDLE_IDENTIFIER "github.com/zeux/pugixml"
```
zeux, line 192: I'm not sure how these get determined by why did this change?

CMakeLists.txt
```diff
@@ -135,7 +141,8 @@ endif()
 
 if (NOT BUILD_SHARED_LIBS OR PUGIXML_BUILD_SHARED_AND_STATIC_LIBS)
   add_library(pugixml-static STATIC
```
zeux, line 143: Why did this change?

Treata11: Is there another way to include the headers in the built library?

CMakeLists.txt
Treata11: I updated the identifiers match the real URL (It'll be included in the `.plist` of the framework.
And `PUBLIC_HEADER` has to be specified. I overlooked it in the previous PR.

CMakeLists.txt
Treata11: `target_include_directories` could be used instead... There isn't much of a difference, if any.

zeux: Static libraries don't include header files, so I'm not sure what this is changing. I'll need to look at how other projects do xcframework setup but it's surprising that any change is necessary here, beyond the change for  framework installation.

zeux: Also worth noting I guess is that xcframework files are dynamic libraries... I'll need to test this fully to make sure I understand what is going on here.

Treata11: > Also worth noting I guess is that xcframework files are dynamic libraries...

`xcframework` are just a bunch of built `framework`s bundled in a single package using XCode-CLI.


>  I'll need to test this fully to make sure I understand what is going on here.

Of course, I actually appreciate that. These links might be helpful:
https://cmake.org/cmake/help/latest/prop_tgt/FRAMEWORK.html#prop_tgt:FRAMEWORK

https://github.com/AcademySoftwareFoundation/Imath/pull/461#issuecomment-2581658965

Treata11:
Hi @zeux,

I think that adding the public headers as static-lib's source, would not interfere with the build process.
I tested regular builds (for macOS) and it was not affected.


Here's a reference that I found, that wasn't committed by me:
https://github.com/libsdl-org/SDL/blob/303fd5ed5a00e3dedaa8fd25b07787f9d5e0e5b5/CMakeLists.txt#L3152C1-L3152C98

zeux:
Sorry for the delay, getting around to this. I understand the need to include the headers now; could you update the other target definition (for shared library) so that these are consistent?

Also, this change changes the framework bundle identifier. I don't have a problem with the old or new name, but why did this change - is there a specific standard convention that this is trying to follow?

Finally, it would be nice to remove the change to `.gitignore` at least as far as `_build` folder is concerned, this name seems specific to a local workflow. `.DS_Store` could stay I guess as it's a common macOS annoyance :)

Treata11:
> Sorry for the delay, getting around to this. I understand the need to include the headers now; could you update the other target definition (for shared library) so that these are consistent?

Sure thing.

> Also, this change changes the framework bundle identifier. I don't have a problem with the old or new name, but why did this change - is there a specific standard convention that this is trying to follow?

The convention is to use reversed domains as bundle id, like:
`com.github.zeux.pugixml` or the web address. 
It only needs to be unique. I simply used the project's Github URL.

> Finally, it would be nice to remove the change to .gitignore at least as far as _build folder is concerned, this name seems specific to a local workflow. .DS_Store could stay I guess as it's a common macOS annoyance :)

Fair enough.

zeux:
Thanks!

PR #660, opened 2025-01-25
refactor: use nullptr

e-kwsm:
https://github.com/zeux/pugixml/blob/9d7fcbf7410e20720b093f9bb3d81dfe6500d30d/CMakeLists.txt#L71-L79

C++11 or later is necessary, so `nullptr` must be available.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,14 +24,12 @@ jobs:
       run: |
         export CXX=${{matrix.compiler}}
         make test cxxstd=c++11 defines=${{matrix.defines}} config=release -j2
-        make test cxxstd=c++98 defines=${{matrix.defines}} config=debug -j2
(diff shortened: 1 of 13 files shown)


e-kwsm:
Thank you for your review. I understand your point.
But `0`/`NULL`, instead of `PUGIXML_NULL`, are used for nullptr, which is fixed in d74511c8b4fae7d78ffad61119e4661d7ca9efcc.

Would you reopen this? I will revert the other commits.

zeux:
> But 0/NULL, instead of PUGIXML_NULL, are used for nullptr, which is fixed in https://github.com/zeux/pugixml/commit/d74511c8b4fae7d78ffad61119e4661d7ca9efcc.

This is by design, you can refer to other PRs around NULL replacements. I'm not interested in further changes in this area.

PR #661, opened 2025-01-25
Modernize use override

e-kwsm:
https://github.com/zeux/pugixml/blob/9d7fcbf7410e20720b093f9bb3d81dfe6500d30d/CMakeLists.txt#L71-L79

C++11 or later is necessary, so `override` must be available.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,14 +24,12 @@ jobs:
       run: |
         export CXX=${{matrix.compiler}}
         make test cxxstd=c++11 defines=${{matrix.defines}} config=release -j2
-        make test cxxstd=c++98 defines=${{matrix.defines}} config=debug -j2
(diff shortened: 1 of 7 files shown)


zeux:
There seems to be some confusion in the two PRs you opened; to save time, I'll just note here that neither of them will be merged. pugixml documents its portability guarantees here https://pugixml.org/docs/manual.html#install.portability - it does not require C++11 and must be compatible with earlier versions of C++ and compilers. CMake *by default* uses C++11 to make sure users get access to C++11 (and now 17) features automatically, but you can override that with CMAKE_CXX_STANDARD as well.

e-kwsm:
Thank you for your review. I understand your point.
But `virtual` is unnecessary for overloaded function, which is fixed in 081e8a32279d4cf62798811957b762f70ea7df80.

Would you reopen this? I will revert the other commits.

zeux:
Feel free to resubmit just that commit as a separate PR.

PR #669, opened 2025-03-29
Silence compiler warnings (-Wall) about switch-case statements not handled

MortenMacFly:
PugiXML is absolutely fantastic, but...

...when you compile it in a large code base where it gets included a few times and you have compiler warnings enabled to the extend to warn about missing switch-case statements, a lot warnings appear for PugiXML cluttering the compiler log.

This patch adds all missing cases as "fall-through" such that 1.) code-checker tools identify it as none-error and 2.) the compiler is silent and "happy".

Hope you like it...

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -4293,6 +4293,9 @@ PUGI_IMPL_NS_BEGIN
 				writer.write('>');
 				break;
 
+			case node_document: // fall-through
+			case node_element:  // fall-through
+			case node_null:     // fall-through
 			default:
 				assert(false && "Invalid node type"); // unreachable
 		}
@@ -6893,6 +6896,7 @@ namespace pugi
 		case node_doctype:
 			return _root->value && (_root->header & impl::xml_memory_page_value_allocated_or_shared_mask) == 0 ? _root->value - doc.buffer : -1;
 
+		case node_null: // fall-through
 		default:
 			assert(false && "Invalid node type"); // unreachable
 			return -1;
@@ -8554,6 +8558,9 @@ PUGI_IMPL_NS_BEGIN
 				return result;
 			}
 
+			case node_declaration: // fall-through
+			case node_doctype:     // fall-through
+			case node_null:        // fall-through
 			default:
 				return xpath_string();
 			}
@@ -9277,6 +9284,7 @@ PUGI_IMPL_NS_BEGIN
 		case xpath_type_boolean:
 			return new_xpath_variable<xpath_variable_boolean>(name);
 
+		case xpath_type_none: // fall-through
 		default:
 			return NULL;
 		}
@@ -9308,6 +9316,7 @@ PUGI_IMPL_NS_BEGIN
 			delete_xpath_variable(static_cast<xpath_variable_boolean*>(var));
 			break;
 
+		case xpath_type_none: // fall-through
 		default:
 			assert(false && "Invalid variable type"); // unreachable
 		}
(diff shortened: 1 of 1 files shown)


zeux:
Which compiler are you using with what settings? pugixml builds in CI using gcc/clang with -Wall -Wextra and MSVC with /W4 and this has not been a problem before.

MortenMacFly:
> Which compiler are you using with what settings? pugixml builds in CI using gcc/clang with -Wall -Wextra and MSVC with /W4 and this has not been a problem before.

I am using MinGW/WinLibs (https://winlibs.com/), GCC 14.2.0 (with POSIX threads) + LLVM/Clang/LLD/LLDB 19.1.7 + MinGW-w64 12.0.0 UCRT - release 3.

I am using PugiXML is mission critical S/W that is compiled with all warnings on by default. These are:

-Wnon-virtual-dtor -Wshadow -Winit-self -Wredundant-decls -Wcast-align -Wundef -Wfloat-equal -Wunreachable-code -Wmissing-declarations -Wmissing-include-dirs -Wswitch-enum -Wswitch-default -Wmain -Wextra -Wall

One of these will cause the warnings. I guess its either: -Wswitch-default  or -Wextra.

MortenMacFly:
> #pragma GCC diagnostic ignored \"-Wswitch-enum\

OK, feel free to do this. However, please be informed (just FYI): For safety critical S/W, which is my daily business, this solution would not be accepted.

If course, PugiXML is probably not targeting such special environments anyway, so I am happy with this solution... Thank you! :-)

zeux:
> this solution would not be accepted
> not targeting such special environments

Since adding additional `case` statements does not change the generated code, and does not meaningfully improve anything unless you actually have audited every single line added, it does not seem like this is solving anything in particular. If you'd like to make an argument that one of these case statements needs to be handled explicitly, I'd be interested in the reasoning.

As far as I know, pugixml is used in some safety critical deployments. I don't know if the code is modified in any way.

Regardless, if you are happy with your pugixml version adding `case` statements to work around this, this is fine and this PR could just be closed.

MortenMacFly:
> As far as I know, pugixml is used in some safety critical deployments. I don't know if the code is modified in any way.

In safety critical software its not only about the binary, but more important also about the traceability, clarity and to avoid any ambiguity.

You solution now introduces: 1.) ambiguity (i.e. did you either forget or explicitly decided _not_ handle the missing cases?); 2.) the danger by disabling these warnings to actually miss something that would be important in the future; 3.) also, its not documented why you disable warnings which it would need to be in safety critical world (even including a full report signed by 2 independent people for authorities for clarification because you make an explicit exception here).

I could elaborate more on this, if you wish, but probably not here... ;-)

But again: Its absolutely fine the way it is now for "the public". The way I am using it is not for the purpose of certifiable S/W!

zeux:
Sure, these arguments are fundamentally stylistical in nature. A different school of thought uses `default` to resolve the supposed ambiguity, which is more aligned with the expectations placed by gcc/clang on their code - there's a reason why Wswitch-enum is not part of Wall/Wextra on either compiler. The problems of this approach are demonstrated fairly clearly in XPath, where your change makes the parser logic, I would say, unambiguously more verbose and no safer.

Anyway, I'm not interested in arguing this, because there's no way to construct a quantifiable argument here on either side. I will note that it looks like only 4 lines of the patch are needed to silence the warning in the core library - all the rest of the patch is in XPath code which can be disabled with PUGIXML_NO_XPATH if you don't need it (which would also improve compile times, as you seem to be using the header mode since otherwise you'd only get the warnings in `pugixml.cpp`). If that's sufficient then I can merge a separate PR that only adds these 4 lines; otherwise it's probably better to keep the code as is.

PR #670, opened 2025-04-19
String length tracking for string_view.

Hurkus:
### Additions
- `xml_attribute_struct` and `xml_node_struct` now track length of name/value with `name_size` and `value_size` properties. String_views obtainable with:
  - `name_sv()`
  - `value_sv()`
- `strcpy_insitu` and friends perform less strlen when copying strings since lengths are known.
- Since the parser now tracks length of parsed strings, merging pcdata performs **2 strlen less per merge**!
- Added a bunch of string_view checks to tests for covering the new string_views and 1 additional attribute value test `parse_attribute_escapes`.

This addition allows for further upgrades to other functions for retrieving child nodes/attributes etc.
Please let me know what you think of this.

### Question
Additionally I'd like to ask about the behaviour tested in `dom_node_anonymous_child` where the name of a child node is deleted and then searched with an empty string_view. It returns a null node. With the new length tracking, nodes with empty names or names that contain a '\0' in the middle (set by the user, since the docs say that it's the users problem for setting invalid names) could be successfully retrieved with appropriate string_views (but not with regular cstrings). I haven't found any definition for correct behaviour.

It would be very convenient if string_views can retrieve nodes with no names or names that have \0 chars as content (allowed by the `name_size` etc.) I've already done some work on this for a future pull request and I'd say it would make things a lot simpler. Let me know if this would be acceptable.

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -3326,7 +3326,7 @@ PUGI_IMPL_NS_BEGIN
 			xml_node_struct* cursor = root;
 			char_t* mark = s;
 			char_t* merged_pcdata = s;
-			size_t* merged_pcdata_len = nullptr;
+			size_t* merged_pcdata_len = NULL;
 
 			while (*s != 0)
 			{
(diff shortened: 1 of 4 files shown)


zeux:
Thanks for the PR! A similar idea has been floated several times with the past, with or without string view. My opinion then, as well as my opinion now, is that this is not a tradeoff I want to make: it increases the size of nodes and attributes significantly, which not only increases the memory consumption but also cache footprint which will invariably come with some sacrifices in performance. XML does not allow using any content with internal `\0` anywhere in the tree, so this does not unlock any previously impossible uses as well. Finally, in compact mode, even modest additions would significantly inflate the tree size, and not doing that in compact mode means we still have the old path without internal `\0` with potential differences in behavior. So I do not plan to merge this.

PR #706, opened 2026-04-20
Fixed Clang -Wextra-semi-stmt warnings

seanm:
Use do-while(0) idiom in several macros to enforce need for trailing semi-colon.


zeux:
Thanks!

PR #708, opened 2026-04-27
cmake: add option to enable std::string_view

nolange:
there was no way to build without STL (more like without iostream), but keeping std::string_view.

its a bit oddly named (`_HAS_`), and IMHO it should be independent whether STL is enabled, but this is
the simplest fix


zeux:
`PUGIXML_NO_STL` disables all uses of STL. This includes iostreams, std::string, iterator tags and std::string_view. So the current implementation is consistent in that sense.

The fact that you can enable `HAS_STRING_VIEW` together with `NO_STL` is pretty accidental due to how the macros are checked; that said it's probably fine to keep as is...

The help text should be adjusted to avoid confusion - this option should almost never be enabled, and I don't want to give users a false impression. Maybe something like "Enable support for std::string_view even if compiler doesn't advertise C++17 or STL usage is disabled".

Out of curiosity what's the use case here?

nolange:
The usecase are freestanding C++ Implementations, this is historically not defined, seems C++26 finally makes an effort: https://cppreference.com/cpp/freestanding

In a nutshell, C++ has lot of simple template only stuff that can be used without trouble on systemd without an OS (eg. newlib without board support.) Stuff like accessing IO or mutex locks dont exist, but you could eg. write an custom interrupt-disable lock and use it with `std::lock`. Similarly `std::string_view` is a really easy to use glue for anything that accepts strings (and can be easily support custom ones)

The c++ library is also static then. Why I singled out iostreams was, that at a glance Pugixml doesnt seem to use much else,
and from experience thats the primary thing to avoid - as soon as you use iostreams the linker will pull in lots of the library, warn that nothing is implemented and bloat the build.

Apart from that, I personally try to avoid iostream, the design always struck me at bad, and there are way better alternatives now: https://en.cppreference.com/cpp/io/print

So what I ideally **would** want:

- not use iostream
- not use locale or other thing I have no use of (pugixml uses strtod, so that would mean using charconv for example)
- still use all those neat C++ abstractions like `std::string_view`
- or in short: have pugixml as "freestanding" static library with a correct config header (#709)

zeux:
Oh I just noticed this is a no-op option. So looks like this PR was not tested? I'm going to revert this.

PR #709, opened 2026-04-27
cmake: install a fitting config header

nolange:
use a header with the correct defines, so
the defines fit to the compiled library.

I have to use this when creating a sysroot / sdk for later builds. the options are currently just discarded.

Ideally the build would use the header too instead of defines, but there is no clean way,
as `#include "pugiconfig.hpp"` will pretty consistentely prefer the file next to it.
I could change this to always create from CMake, but I guess that breaks all the other build systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ separate_arguments(PUGIXML_BUILD_DEFINES)
 
 # Public headers
 set(PUGIXML_HEADERS
-    ${PROJECT_SOURCE_DIR}/src/pugiconfig.hpp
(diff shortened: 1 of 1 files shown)


zeux:
This can't be unconditional. It needs to be gated under `PUGIXML_INSTALL_CONFIG` option; I'm not sure if it's better to change `PUGIXML_HEADERS` or just adjust the installation code. Also at this point I wonder if the sysroot build even needs to use CMake; it looks like your use case requires a very custom configuration and a custom config file.

nolange:
> it looks like your use case requires a very custom configuration and a custom config file.

I would argue packages normally install a fitting config header to the library. That would be a bigger change and should cover all build systems. Converted this to draft. feel free to close it.

PR #710, opened 2026-05-12
Use charconv for numeric parsing printing

nolange:
This hooks up https://en.cppreference.com/cpp/utility/from_chars and https://en.cppreference.com/cpp/utility/to_chars,
it will trigger automatically if code is compiled for C++17 and a C++ prodiving *float* support for these functions is available.

patch should be straight forward, the calls into stdio functions are replaced, charconv is independend from locale,
but stops at the first character that cant be parsed, including whitespace - so those must be manually skipped.

**The first commit is compatible with the test framwork**, the second tries to naively use charconv in convert_number_to_string.
I dont understand why another function is necessary to ectract mantissa / exp an construct an own kind
of double to string conversion.

trying to dig into the xpath stuff gave me headaches. I would want to merge only the first commit, but want to clear up if we can fix / replace the convert_number_to_string function too. I believe the string representation is different, but the parsed float value would be identical.

```
Test xpath_xalan_string_9 failed: "string(.0000123456789)" does not evaluate to "0.0000123456789" in context c at pugixml/tests/test_xpath_xalan_2.cpp:336
Test xpath_xalan_string_8 failed: "string(9876543.21012345)" does not evaluate to "9876543.21012345" in context c at pugixml/tests/test_xpath_xalan_2.cpp:303
Test xpath_xalan_string_6 failed: "string(1234567)" does not evaluate to "1234567" in context c at pugixml/tests/test_xpath_xalan_2.cpp:197
Test xpath_xalan_math_9 failed: "string(number('0.0000000000001'))" does not evaluate to "0.0000000000001" in context c at pugixml/tests/test_xpath_xalan_1.cpp:395
Test xpath_string_string failed: "string(1234567)" does not evaluate to "1234567" in context c at pugixml/tests/test_xpath_functions.cpp:280
Test xpath_out_of_memory_evaluate_number_to_string failed: q.evaluate_string(0, 0, xml_node()) == 1 is false at pugixml/tests/test_xpath.cpp:578
Test xpath_denorm_numbers failed: query.c_str() does not evaluate to "0.0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000009999987484955998" in context xml_node() at pugixml/tests/test_xpath.cpp:275
Test xpath_long_numbers_stringize failed: test_xpath_string_prefix(c, str_flt_max, str_flt_max, 15) is false at pugixml/tests/test_xpath.cpp:252
FAILURE: 8 out of 983 tests failed.
```

### DIFF
--- a/docs/manual.adoc
+++ b/docs/manual.adoc
@@ -228,6 +228,8 @@ pugixml uses several defines to control the compilation process. There are two w
 
 [[PUGIXML_WCHAR_MODE]]`PUGIXML_WCHAR_MODE` define toggles between UTF-8 style interface (the in-memory text encoding is assumed to be UTF-8, most functions use `char` as character type) and UTF-16/32 style interface (the in-memory text encoding is assumed to be UTF-16/32, depending on `wchar_t` size, most functions use `wchar_t` as character type). See <<dom.unicode>> for more details.
 
+[[PUGIXML_CHARCONV_FLOAT]]`PUGIXML_CHARCONV_FLOAT` will use https://en.cppreference.com/cpp/header/charconv[<charconv>] for floating-point number formatting instead of `<stdio.h>` functions, which requires C++17 and UTF-8 interface. Note that the conversion will then ignore the locale and always act as if the default `C` locale is used.
+
 [[PUGIXML_COMPACT]]`PUGIXML_COMPACT` define activates a different internal representation of document storage that is much more memory efficient for documents with a lot of markup (i.e. nodes and attributes), but is slightly slower to parse and access. For details see <<dom.memory.compact>>.
 
 [[PUGIXML_NO_XPATH]]`PUGIXML_NO_XPATH` define disables XPath. Both XPath interfaces and XPath implementation are excluded from compilation. This option is provided in case you do not need XPath functionality and need to save code space.
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -4807,7 +4807,7 @@ PUGI_IMPL_NS_BEGIN
 	{
 		char buf[128];
 	#ifdef PUGIXML_CHARCONV_FLOAT
-		std::to_chars_result result = std::to_chars(std::begin(buf), std::end(buf), value, std::chars_format::general, precision);
+		std::to_chars_result result = std::to_chars(std::begin(buf), std::end(buf) - 1, value, std::chars_format::general, precision);
 		*result.ptr = '\0';
 	#else
 		PUGI_IMPL_SNPRINTF(buf, "%.*g", precision, double(value));
@@ -4820,7 +4820,7 @@ PUGI_IMPL_NS_BEGIN
 	{
 		char buf[128];
 	#ifdef PUGIXML_CHARCONV_FLOAT
-		std::to_chars_result result = std::to_chars(std::begin(buf), std::end(buf), value, std::chars_format::general, precision);
+		std::to_chars_result result = std::to_chars(std::begin(buf), std::end(buf) - 1, value, std::chars_format::general, precision);
 		*result.ptr = '\0';
 	#else
 		PUGI_IMPL_SNPRINTF(buf, "%.*g", precision, value);
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -35,6 +35,10 @@
 #	include <string>
 #endif
 
+#ifdef PUGIXML_CHARCONV_FLOAT
+#	include <charconv>
+#endif
+
 // For placement new
 #include <new>
 
@@ -4700,6 +4704,14 @@ PUGI_IMPL_NS_BEGIN
 	{
 	#ifdef PUGIXML_WCHAR_MODE
 		return wcstod(value, NULL);
+	#elif defined(PUGIXML_CHARCONV_FLOAT)
+		while (PUGI_IMPL_IS_CHARTYPE(*value, ct_space))
+			value++;
+		if (*value == '+')
+			value++;
+		double result = 0.0;
+		std::from_chars(value, value + strlen(value), result);
+		return result;
 	#else
 		return strtod(value, NULL);
 	#endif
@@ -4709,6 +4721,14 @@ PUGI_IMPL_NS_BEGIN
 	{
 	#ifdef PUGIXML_WCHAR_MODE
 		return static_cast<float>(wcstod(value, NULL));
+	#elif defined(PUGIXML_CHARCONV_FLOAT)
+		while (PUGI_IMPL_IS_CHARTYPE(*value, ct_space))
+			value++;
+		if (*value == '+')
+			value++;
+		float result = 0.0f;
+		std::from_chars(value, value + strlen(value), result);
+		return result;
 	#else
 		return static_cast<float>(strtod(value, NULL));
 	#endif
@@ -4786,8 +4806,12 @@ PUGI_IMPL_NS_BEGIN
 	PUGI_IMPL_FN bool set_value_convert(String& dest, Header& header, uintptr_t header_mask, float value, int precision)
 	{
 		char buf[128];
+	#ifdef PUGIXML_CHARCONV_FLOAT
+		std::to_chars_result result = std::to_chars(std::begin(buf), std::end(buf), value, std::chars_format::general, precision);
+		*result.ptr = '\0';
+	#else
 		PUGI_IMPL_SNPRINTF(buf, "%.*g", precision, double(value));
-
+	#endif
 		return set_value_ascii(dest, header, header_mask, buf);
 	}
 
@@ -4795,7 +4819,12 @@ PUGI_IMPL_NS_BEGIN
 	PUGI_IMPL_FN bool set_value_convert(String& dest, Header& header, uintptr_t header_mask, double value, int precision)
 	{
 		char buf[128];
+	#ifdef PUGIXML_CHARCONV_FLOAT
+		std::to_chars_result result = std::to_chars(std::begin(buf), std::end(buf), value, std::chars_format::general, precision);
+		*result.ptr = '\0';
+	#else
 		PUGI_IMPL_SNPRINTF(buf, "%.*g", precision, value);
+	#endif
 
 		return set_value_ascii(dest, header, header_mask, buf);
 	}
@@ -8862,14 +8891,22 @@ PUGI_IMPL_NS_BEGIN
 #else
 	PUGI_IMPL_FN void convert_number_to_mantissa_exponent(double value, char (&buffer)[32], char** out_mantissa, int* out_exponent)
 	{
+	#ifdef PUGIXML_CHARCONV_FLOAT
+		std::to_chars_result res = std::to_chars(std::begin(buffer), std::end(buffer) - 1, value, std::chars_format::scientific, DBL_DIG);
+		*res.ptr = '\0';
+	#else
 		// get a scientific notation value with IEEE DBL_DIG decimals
 		PUGI_IMPL_SNPRINTF(buffer, "%.*e", DBL_DIG, value);
-
+	#endif
 		// get the exponent (possibly negative)
 		char* exponent_string = strchr(buffer, 'e');
 		assert(exponent_string);
 
-		int exponent = atoi(exponent_string + 1);
+		char *s = exponent_string + 1;
+		bool isneg = *s++ == '-';
+		int exponent = 0;
+		while (*s) exponent = exponent * 10 + (*s++ - '0');
+		exponent = isneg ? -exponent : exponent;
 
 		// extract mantissa string: skip sign
 		char* mantissa = buffer[0] == '-' ? buffer + 1 : buffer;
@@ -8958,9 +8995,6 @@ PUGI_IMPL_NS_BEGIN
 
 	PUGI_IMPL_FN bool check_string_to_number_format(const char_t* string)
 	{
-		// parse leading whitespace
-		while (PUGI_IMPL_IS_CHARTYPE(*string, ct_space)) ++string;
-
 		// parse sign
 		if (*string == '-') ++string;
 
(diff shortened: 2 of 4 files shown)


nolange:
Btw. from_chars would need an "end of buffer" (not necessary end of the parsable value ). I used `s + std::char_traits<char_t>::length(s)`, adding a pointer there thats guaranteed to be in buffer without using strlen would be nice.
could just use `s + 300`, the parsing will stop at `\0`, but strictly standardese would claim this is undefined behaviour.

zeux:
I would be fine with this as an opt-in feature; it can not be enabled automatically as it changes behavior in locale-sensitive code. It should be off by default.

It should only apply to float formatting/parsing; this change applies it in other cases. Only calls to strtod/sprintf-with-%.g should be affected.

Please fix the code style and mixed tab/spaces as well to align with the existing style around it; pugixml doesn't use `auto`, `var{}` initializers et al.

zeux:
Also `charconv` is decidedly part of STL. It looks like the changes try to support NO_STL + CHARCONV_FLOAT; we should not be including `<string>` if NO_STL is defined. I would question if these should even be valid when defined together; similar to `string_view`. May not be critical if `<string>` header dependency is removed.

nolange:
> I would be fine with this as an opt-in feature; it can not be enabled automatically as it changes behavior in locale-sensitive code. It should be off by default.

Ok. should XML parsing be locale specific? given this is a data.driven format I always assumed it should not,

> It should only apply to float formatting/parsing; this change applies it in other cases. Only calls to strtod/sprintf-with-%.g should be affected.

well, replacing sprintf and the stdio stuff it pulls in is kinda the point.

> 
> Please fix the code style and mixed tab/spaces as well to align with the existing style around it; pugixml doesn't use `auto`, `var{}` initializers et al.

ok, I assumed this was for compat reasons, and the charconv stuff is guarded by C++17

nolange:
> Also `charconv` is decidedly part of STL. It looks like the changes try to support NO_STL + CHARCONV_FLOAT; we should not be including `<string>` if NO_STL is defined. I would question if these should even be valid when defined together; similar to `string_view`. May not be critical if `<string>` header dependency is removed.

I use string only to get `std::char_traits`, guess I can try using strlen instead.

Again, I thought this would largely be for dealing with old compilers, not those that advertise C++17.

zeux:
> Ok. should XML parsing be locale specific? given this is a data.driven format I always assumed it should not

It's application specific as to whether the numbers stored in XML are using locale sensitive formatting or not. XPath has stricter requirements so I would not be worried about it if it was just for XPath code, but for general DOM access there are applications that rely on locale specific formatting as well as ones that want the processing to be locale-independent. The DOM nodes store text, so it's really just about the numeric APIs (as_float/etc.)

nolange:
this should now fit your requests.

- removed from_chars where not necessary but replaced all stdio calls.
- renamed the flag to PUGIXML_CHARCONV_CONVERSION as it also replaces one atoi call.
- added variable to CMake
- added asciidoc (did not regen html)
- style fixes

I also tried adding it to ci, but we would need to filter out configurations not supporting C++17:

https://github.com/nolange/pugixml

zeux:
I think `PUGIXML_CHARCONV_FLOAT` was a better name; the goal here is locale independent formatting for floating point literals as far as I'm concerned. The `atoi` call could be left as is, I'm not sure why we need to replace it - if `atoi` is a bad fit there it's better to implement it manually anyhow, as the behavior needs to be a small subset (this is not parsing user input, it's extracting the exponent from the formatted output because XPath requires specific handling of literals with lots of decimal/significant digits). But I'd be fine with keeping the relevant portion of the patch as is too, as `atoi` there is just part of the number formatting so it fits the higher level theme of "change floating point printing/parsing".

Additionally, I think it's important that this works in `wchar_t` mode too. You can use `impl::strlength` to replace strlen in a way that will work with either version.

It's fine to not add this to CI for now, I can take a look at that separately after a version of this gets merged.

src/pugixml.cpp
```diff
@@ -8858,14 +8887,29 @@ PUGI_IMPL_NS_BEGIN
...
 		// get the exponent (possibly negative)
 		char* exponent_string = strchr(buffer, 'e');
 		assert(exponent_string);
 
+	#ifdef PUGIXML_CHARCONV_CONVERSION
+		const char *s = exponent_string + 1;
+		bool isneg = *s == '-';
+		while (*s++ == '0')
+			;
+		int exponent;
```
zeux:
```suggestion
		bool isneg = *s++ == '-'; // skip +/- sign
		int exponent = 0;
```
This took me a minute because I couldn't see why this works for positive exponents. If we *are* removing `atoi` we should do it like this at least.

nolange: i dont think I fully understand that function. the problem is that the string will be something like `e-01` and the issue is the zero. the from_chars function does not permit leading zeros, the sign is not the problem.

What do you think of using this (independ of this PR, just completely replace atoi):

`int exponent = string_to_integer(exponent_string + 1, -INT_MAX, INT_MAX);`

using INT_MIN instead of  -INT_MAX will result in and overflow btw, there seems to be a bug in the function.

nolange:
> Additionally, I think it's important that this works in `wchar_t` mode too. You can use `impl::strlength` to replace strlen in a way that will work with either version.

from_chars/to_chars only works with char, there is no wchar_t implementation, so using strlen is fine there.

if you permit including `<string>` or `<string_view>` together with `<charconv>` i could use `std::char_traits` as before, or construct a `std::string_view` - in both variants the library will chose the best function, potentially a compiler builtin for strlen.

(ideally we would pass a valid end-of-buffer instead of using strlen, but I get this is a big change)

so whats your pick?

src/pugixml.cpp
zeux: > from_chars function does not permit leading zeros

It does, and your code doesn't skip them correctly - the `while` loop never runs as the first character is always a `+` or `-`, but it does advance `s` by 1. The sign *is* a problem because exponent starts with `+`.

As I mentioned before you don't need to handle overflow here; this is parsing the exponent that was just printed and that has a specified and constrained format. If you wanted a custom implementation this could literally just be:

```c++
bool isneg = *s++ == '-';
int exponent = 0;
while (*s) exponent = exponent * 10 + (*s++ - '0');
exponent = isneg ? -exponent : exponent;
```

zeux:
> from_chars/to_chars only works with char, there is no wchar_t implementation

Oh, apologies, I was assuming they would just work. That's problematic, especially for parsing because the input string is unbounded so to handle this we'd need to convert it on heap. Okay... in that case I think it's fine to keep this incompatible with wchar mode.

docs/manual.adoc
```diff
@@ -228,6 +228,8 @@ pugixml uses several defines to control the compilation process. There are two w
 
 [[PUGIXML_WCHAR_MODE]]`PUGIXML_WCHAR_MODE` define toggles between UTF-8 style interface (the in-memory text encoding is assumed to be UTF-8, most functions use `char` as character type) and UTF-16/32 style interface (the in-memory text encoding is assumed to be UTF-16/32, depending on `wchar_t` size, most functions use `wchar_t` as character type). See <<dom.unicode>> for more details.
 
+[[PUGIXML_CHARCONV_CONVERSION]]`PUGIXML_CHARCONV_CONVERSION` will use number formating via https://en.cppreference.com/cpp/header/charconv[<charconv>] instead of `<stdio.h>` functions, this requires C++17 and UTF-8 interface. Note that the conversion will then ignore the locale and always als like the default `C` locale is used.
```
zeux:
```suggestion
[[PUGIXML_CHARCONV_FLOAT]]`PUGIXML_CHARCONV_FLOAT` will use https://en.cppreference.com/cpp/header/charconv[<charconv>] for floating-point number formatting instead of `<stdio.h>` functions, which requires C++17 and UTF-8 interface. Note that the conversion will then ignore the locale and always act as if the default `C` locale is used.
```

src/pugixml.cpp
nolange: I totally dont want custom implementation, my motivation is to avoid the printf/scanf package (atoi is part of it) as this will pull in pretty much everything from a static C library. the options are:

- use string_to_integer
- use from_chars (but obviously only if the option is enabled)
- some custom stuff.

I would lean on using string_to_integer, as this is already available and used and doesnt depend on options. perhaps only grab the sign and use the unsigned version that is already used elsewhere to save some code.

docs/manual.adoc
nolange: done

nolange:
- I swapped atoi with string_to_integer (couldve been a separate PR).
- docs updated with your changes

Would just need to rebuild html.

zeux:
Thanks! I can update HTML docs after merging this, will just wait for the builds to pass.

nolange:
forgot to fix some `auto`  variables. should be the final version now

src/pugixml.cpp
```diff
@@ -4786,16 +4806,25 @@ PUGI_IMPL_NS_BEGIN
 	PUGI_IMPL_FN bool set_value_convert(String& dest, Header& header, uintptr_t header_mask, float value, int precision)
 	{
 		char buf[128];
+	#ifdef PUGIXML_CHARCONV_FLOAT
+		std::to_chars_result result = std::to_chars(std::begin(buf), std::end(buf), value, std::chars_format::general, precision);
```
zeux:
```suggestion
		std::to_chars_result result = std::to_chars(std::begin(buf), std::end(buf) - 1, value, std::chars_format::general, precision);
```

src/pugixml.cpp
```diff
@@ -4786,16 +4806,25 @@ PUGI_IMPL_NS_BEGIN
...
+	#endif
 		return set_value_ascii(dest, header, header_mask, buf);
 	}
 
 	template <typename String, typename Header>
 	PUGI_IMPL_FN bool set_value_convert(String& dest, Header& header, uintptr_t header_mask, double value, int precision)
 	{
 		char buf[128];
+	#ifdef PUGIXML_CHARCONV_FLOAT
+		std::to_chars_result result = std::to_chars(std::begin(buf), std::end(buf), value, std::chars_format::general, precision);
```
zeux:
```suggestion
		std::to_chars_result result = std::to_chars(std::begin(buf), std::end(buf) - 1, value, std::chars_format::general, precision);
```

zeux (commented):
Two off by one issues, will apply the patch.

zeux:
Interestingly, `from_chars` behavior for overflow seems to diverge from the commonly accepted practice (overflow to infinity); it instead returns an error code but does not communicate the input sign at all. I'm not sure what the XPath requirements are on this - the specification does not say one way or the other. I'll take a look at this after merging this and might submit some changes alongside docs updates in the future but this should be a good baseline.

zeux:
Oh, it's actually worse: `from_chars` will report out of range on over- or under- flow. This patch currently reports `0` which is more-correct for underflow and less-correct for overflow. https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2827r1.html is a great summary of the problem. It sounds like attempts for working around this may be unstable for future standard revisions depending on the resolution for this... so I'll leave this as is for now I suppose, the feature is opt-in anyhow.

nolange:
yes, the functions are more like building blocks doing the least amount of work. 
i think the as_float/as_double need to be adopted too. i guess inf/nan strings are not supported either.

zeux:
Per https://isocpp.org/files/papers/P4168R0.html#limited-design-options, actually libc++ and MSVC already diverge from the standard in the direction that seems reasonable to me. So in that sense, the standard has questionable behavior that libstdc++ implements faithfully but MSVC/libc++ take a more defensible position in practice. The code in this change will thus work fine with implementation specific dependencies; so be it.

nolange:
im gonna have to to look at it later in depth, but i think the spec is that underflow will either result in 0 or the smallest value with correct sign. only overflow will give you an error. so the first article is not completely right - there is no information about underflow, but the error flag always means overflow.

zeux:
No, the standard wording is specifically:

> If the parsed value is not in the range representable by the type of value, value is unmodified and the member ec of the return value is equal
> to errc::result_out_of_range. Otherwise, value is set to the parsed value, after rounding according to
round_to_nearest (21.3.3.1), and the member ec is value-initialized.

I've verified that on libstdc++, strings "1e-400" and "1e+400" both result in the same behavior: output not modified, result error code out_of_range. And that on libc++, result is out_of_range in both cases, but the output value is modified to be 0 for 1e-400 and +inf for 1e+400.

So again no change required here, but the behavior of this code would be dependent on the STL implementation.

nolange:
I waited for ages for charconv to end up in Clang, upon further reading this is a mess. Thanks for bringing this to my attention

I will try adding handling for libstdc++ only in another MR (libc++ and MSVC are already correct according to [P4168R0](https://wg21.link/P4168R0)).

zeux:
> adding handling for libstdc++ only in another MR

No - please don't. I think status quo is fine and it has to wait until C++ WG & libstdc++ resolve the final behavior here. In the meantime, overflown numbers will return 0 on libstdc++. I'm not willing to maintain divergent STL specific implementations.

nolange:
i was thinking about adding a function thats called when result_out_of_range is set - only on libstdc++, then see if the string had a negative exp or is in the form `0+.`. handfull of char searches, not a full blown replacement, maybe 15 LOC postprocessing in the error case. 
I need to get my head around the implications anyway.

ill post it here so I dont burn a MR you will just close

nolange:
first draft, will be called if libstdc++ is used and `ec == std::errc::result_out_of_range`. Gonna need some days testing, but I think this covers all underflow cases (well... outside of hex-floats)

``` c++
	float fixup_from_chars(const char *begin, const char *end) {
		bool is_neg = *begin == '-';
		bool is_underflow = false;

		std::string_view s(begin + (is_neg ? 1 : 0), end - begin - (is_neg ? 1 : 0));
		std::string_view::size_type exp_ind = s.find_first_of("eE");

		if (exp_ind != std::string_view::npos) {
			// Scientific notation, test if exponent is negative
			is_underflow = s.find('-', exp_ind) != std::string_view::npos;
		} else {
			// Fixed-point Notation, test if the whole part is zero
			std::string_view::size_type dot_ind = s.find('.');
                        if (dot_ind != std::string_view::npos) {
                        	// test if sting before is all zero (or empty)
                        	is_underflow = s.substr(0, dot_ind).find_first_not_of('0') == std::string_view::npos;
                        }
		}
		return is_underflow ? (is_neg ? -0.0f : 0.0f) : (is_neg ? -HUGE_VALF : HUGE_VALF);
	}
```

nolange:
Fixed version in https://github.com/nolange/pugixml/tree/fixup_charconv_libstdcpp

Still want no PR?

Issue #714, opened 2026-05-15
Heap-buffer-overflow in `pugi::xml_node::first_element_by_path` when `delimiter == '\0'`


FalconLi:
## Summary
Heap-buffer-overflow in `pugi::xml_node::first_element_by_path` function when delimiter parameter is null character.

## Version
```
$ git describe --tags
latest-35-g0595296
```

## Description
The crash occurs in `xml_node::first_element_by_path` when the caller passes `'\0'` (null character) as the `delimiter` argument. The function uses three loops that compare characters against `delimiter`, and one of them is also expected to terminate at the string's null terminator. When `delimiter` itself equals `'\0'`, this dual role collapses and the loop reads past the end of the buffer.

Specifically, after determining the current path segment, the function executes:

```cpp
const char_t* next_segment = path_segment_end;          // points at '\0'
while (*next_segment == delimiter) ++next_segment;      // delimiter == '\0'
```

Because `path_segment_end` was advanced until `*path_segment_end == '\0'` (i.e., the string terminator), `next_segment` initially points at the trailing `'\0'` of `path_`. With `delimiter == '\0'`, the condition `*next_segment == delimiter` is true at the terminator, so the loop keeps incrementing `next_segment` past the end of the heap-allocated path string, causing the heap-buffer-overflow detected by ASan.

The root cause is that the function's design implicitly assumes `delimiter` is a non-null, non-terminator character, but it does not validate this precondition.

## PoC Code
```cpp
#include "pugixml.hpp"
#include <cstdint>
#include <string>
#include <vector>

extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
    
    std::vector<std::string> paths = {
        "level1",
        "level1.level2",
        "level1.level2.level3",
        "level1.level3",
        "level1.level2.level4",
        "",
        "level1."
    };
    char path_delimiter = '\0';

    // Create XML document
    pugi::xml_document doc;
    pugi::xml_node start = doc.append_child("root");
    pugi::xml_node level1 = start.append_child("level1");
    level1.append_attribute("id") = "1";
    level1.append_attribute("type") = "parent";
    level1.text().set("Level 1 text");
    pugi::xml_node level2a = level1.append_child("level2");
    level2a.append_attribute("id") = "2a";
    level2a.append_attribute("category") = "A";
    level2a.text().set("Level 2A text");
    pugi::xml_node level2b = level1.append_child("level2");
    level2b.append_attribute("id") = "2b";
[...]

PR #715, opened 2026-05-15
Reject null delimiter in first_element_by_path to prevent heap-buffer-overflow

FalconLi:
Fixes the issue described in #714

Heap-buffer-overflow occurs in `pugi::xml_node::first_element_by_path` function when delimiter parameter is null character.

The fix rejects a null-character delimiter at the top of `first_element_by_path` by returning an empty `xml_node`, preventing the segment-skipping loops from advancing past the string's null terminator and causing a heap-buffer-overflow.

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -6786,6 +6786,8 @@ namespace pugi
 
 	PUGI_IMPL_FN xml_node xml_node::first_element_by_path(const char_t* path_, char_t delimiter) const
 	{
+		if (delimiter == 0) return xml_node();
+		
 		xml_node context = path_[0] == delimiter ? root() : *this;
 
 		if (!context._root) return xml_node();


zeux:
This should be an assert.

FalconLi:
Dear pugixml developers,

We have submitted this issue and the PR as we believe that `'\0'` is a valid delimiter in C++. However, this project does not consider the `'\0'` delimiter situation (the documentation also does not state that `'\0'` is an invalid delimiter for this project) and an address sanitizer crash occurs due to unchecked `'\0'` delimiter in the `first_element_by_path()` function. 

We will give the following example to demonstrate that `'\0'` could be a valid delimiter in C++:
```cpp
#include <iostream>
#include <string>
#include <sstream>

int main() {
    std::string data("Hello\0World\0C++", 15);
    std::stringstream ss(data);
    std::string token;

    while (std::getline(ss, token, '\0')) {
        std::cout << "Token: " << token << " (length: " << token.length() << ")\n";
    }

    return 0;
}
```
The output of the previous example is as follows:
```
$ ./example
Token: Hello (length: 5)
Token: World (length: 5)
Token: C++ (length: 3)
```
This example shows that `'\0'` can be used as a delimiter in C++. Therefore, this situation needs to be handled in the `first_element_by_path()` function to prevent the address sanitizer crash as described in the issue.

Our current change in the PR is following the logic of the project:
* the `'\0'` delimiter is invalid for this project
* invalid situations return an empty node using `return xml_node();`

So our change was
```diff
 PUGI_IMPL_FN xml_node xml_node::first_element_by_path(const char_t* path_, char_t delimiter) const
 {
+	if (delimiter == 0) return xml_node();
+
 	xml_node context = path_[0] == delimiter ? root() : *this;
```

> This should be an assert.

Based on my understanding, you want the `'\0'` delimiter situation be an assert instead of returning an empty node.
So I am going to make the following change if you affirm my understanding.
```diff
 PUGI_IMPL_FN xml_node xml_node::first_element_by_path(const char_t* path_, char_t delimiter) const
 {
-	if (delimiter == 0) return xml_node();
+	assert(delimiter!=0);

 	xml_node context = path_[0] == delimiter ? root() : *this;
```

Thank you for your reply.

PR #716, opened 2026-05-15
Improve and test std::charconv integration

### LINKED ISSUE #469: parsing float is local dependant
In France, decimal separator is "," and not "."

strtod used in pugixml.cpp uses strtod which is local dependant

possible alternative:
1) use own double parsing code like https://github.com/michael-hartmann/parsefloat
2) use std::from_chars is c++17 is detected

### LINKED ISSUE #568: strtod is locale dependent, so as xml_attribute::as_double and get_value_double
There're 2 main decimal separators: comma and dot:

https://en.wikipedia.org/wiki/Decimal_separator#Countries_using_decimal_comma

If you save XML on e.g. UK locale computer and send that to a person in PL, then double values are gonna be converted to INT

char decimalComma[] = "123,123";
char decimalPoint[] = "123.123";

qDebug() << "--------" << strtod(decimalComma, 0) << strtod(decimalPoint, 0);

output depends on a locale:
- in PL, the output will be "123.123", "123"
- in UK, the output will be "123", "123.123"

it would be great if pugixml would handle that internally. We use QByteArrayView::toDouble() as a bypass for now.

zeux:
This is based on #710 and adds GHA integration as well as fixing the error handling for `to_chars`: on error, it would store the *end* of the buffer in `result.ptr`, which would produce an uninitialized but zero terminated string. We now return false on error (which should only be reachable if abnormally large precision is requested externally).

Note that for correct `from_chars` behavior on out of range floats, we prefer libc++/MSVC STL behavior of setting the output value to zero/infinity; C++ standard specifies broken behavior that libstdc++ implements, in which case we'd return 0 (which is okay on underflow but obviously not ideal for overflow - it's impossible to distinguish these when using libstdc++). See https://isocpp.org/files/papers/P4168R0.html for details.

Since for locale-independent behavior you can now compile pugixml with `PUGIXML_CHARCONV_FLOAT`, this also:

- Closes #470
- Fixes #469
- Fixes #568

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,26 +12,34 @@ jobs:
       matrix:
         os: [ubuntu, macos]
         compiler: [g++, clang++]
-        defines: [standard, PUGIXML_WCHAR_MODE, PUGIXML_COMPACT, PUGIXML_NO_EXCEPTIONS]
+        defines: [standard, PUGIXML_WCHAR_MODE, PUGIXML_COMPACT, PUGIXML_NO_EXCEPTIONS, PUGIXML_CHARCONV_FLOAT]
         exclude:
           - os: macos
             compiler: g++
+          - os: macos
+            defines: PUGIXML_CHARCONV_FLOAT # std::charconv is not supported until Xcode 26
     name: ${{matrix.os}} (${{matrix.compiler}}, ${{matrix.defines}})
     runs-on: ${{matrix.os}}-latest
     steps:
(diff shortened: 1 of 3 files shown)


zeux:
Note: on macOS, this seems to be available since Xcode 26 by default; GHA is currently on Xcode 16.4. So the build configuration is disabled for macOS builds until that changes.

Also intentionally excluding this new mode from AppVeyor, as it's already struggling.

PR #717, opened 2026-05-16
Add VS2026 projects

zeux:
Closes #693

The new project files were copied from VS2022 and upgraded to the new toolchain with no further changes. This also adds the new version to the PowerShell script but doesn't integrate it with Appveyor yet - Appveyor doesn't have a VS2026 image yet. See https://github.com/appveyor/ci/issues/3985.


zeux:
VS2013 & VS2015 are also officially out of support per https://learn.microsoft.com/en-us/visualstudio/releases/2026/servicing-vs so this will be a good opportunity to remove them from NuGet packages too.

zeux:
Will split Appveyor changes into a separate PR.

PR #718, opened 2026-05-18
Update AppVeyor configuration for VS2026

zeux:
- Add VS2026 builds to AppVeyor configuration
- Remove VS2013/VS2015 from AppVeyor NuGet packages

VS2013/2015 are out of support according to https://learn.microsoft.com/en-us/visualstudio/releases/2026/servicing-vs.

Note that for now we still build and test pugixml itself for the older VS versions; the removals here only apply to the NuGet package.

This won't build until https://github.com/appveyor/ci/issues/3985 is fixed.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,7 @@ branches:
     - master
 
 image:
+  - Visual Studio 2026
   - Visual Studio 2022
   - Visual Studio 2019
   - Visual Studio 2017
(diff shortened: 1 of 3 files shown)


zeux:
Good news: AppVeyor now has a Visual Studio 2026 image!
Bad news: the said image has CMake 4.1.2 which doesn't support VS2026 yet so the build still fails.

PR #719, opened 2026-05-18
Delete xml_attribute::set_value/xml_text::set overloads for char8_t*

### LINKED ISSUE #378: PugiXML incorrectly handles char8_t in C++20

zeux:
When using C++20 in applications that use `const char8_t*`, it's possible to accidentally use attribute/text assignment (`set_value`/`set`/`operator=`) without casting to `const char*`.

Previously, these calls would compile - the pointer would be implicitly cast to `bool`. This is almost definitely a mistake in the code, and we now force the caller to cast by deleting these overloads.

It would be possible to delete `const void*` overloads instead to cover *any* pointers distinct from `char_t`, but this is more likely to break valid code that sets booleans intentionally using the pointer NULLness, such as:

	node.append_attribute("has_data") = data;
	if (data) ...

Fixes #378.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,7 @@ jobs:
         make test cxxstd=c++11 defines=${{matrix.defines}} config=release -j2
         make test cxxstd=c++98 defines=${{matrix.defines}} config=debug -j2
         make test cxxstd=c++17 defines=${{matrix.defines}} config=debug -j2
+        make test cxxstd=c++20 defines=${{matrix.defines}} config=debug -j2
         make test defines=${{matrix.defines}} config=sanitize -j2
     - name: make test charconv
       if: matrix.defines == 'PUGIXML_CHARCONV_FLOAT'
--- a/src/pugixml.hpp
+++ b/src/pugixml.hpp
@@ -493,6 +493,12 @@ namespace pugi
 		xml_attribute& operator=(unsigned long long rhs);
 	#endif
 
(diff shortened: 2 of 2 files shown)


brandl-muc:
Why not implement these functions by reintepret_cast()ing and delegating?

zeux:
I considered it but since no other functions support char8_t, and duplicating everything is not something I plan for, disabling seemed better because it doesn’t give a false impression that pugixml supports char8_t - and merely blocks a potential footgun mentioned in the linked issue.

brandl-muc:
I understand.
I don't even know how proper char8_t support could be implemented.

zeux:
After mulling this over I'm actually not sure I like this, because the same problem exists between wchar_t/char and I can easily see similar mistakes being made there - which this issue doesn't address. Deleting the overload for `const void*` would but I'm worried this will break some existing code with no great way to check. Will close for now and reconsider.

Issue #721, opened 2026-05-27
CRAN (R Programming Language) warning about pugixml.cpp


pachadotdev:
The R checks for the 'redatam' package that vendors 'pugixml' show this for the released 'pugixml' 1.15

https://cran.r-project.org/web/checks/check_results_redatam.html

```
Check Details

Version: 2.3.0
Check: whether package can be installed
Result: WARN 
  Found the following significant warnings:
    vendor/pugixml.cpp:491:66: warning: ‘*other.pugi::impl::{anonymous}::xml_document_struct::pugi::xml_node_struct.pugi::xml_node_struct::header’ is used uninitialized [-Wuninitialized]
  See ‘/home/hornik/tmp/R.check/r-devel-gcc/Work/PKGS/redatam.Rcheck/00install.out’ for details.
  * used C++ compiler: ‘g++-16 (Debian 16.1.0-1) 16.1.0’
Flavor: [r-devel-linux-x86_64-debian-gcc](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-gcc/redatam-00check.html)

Version: 2.3.0
Check: whether package can be installed
Result: WARN 
  Found the following significant warnings:
    vendor/pugixml.cpp:491:66: warning: ‘*other.pugi::impl::{anonymous}::xml_document_struct::pugi::xml_node_struct.pugi::xml_node_struct::header’ is used uninitialized [-Wuninitialized]
  See ‘/home/hornik/tmp/R.check/r-release-gcc/Work/PKGS/redatam.Rcheck/00install.out’ for details.
  * used C++ compiler: ‘g++-16 (Debian 16-20260425-1) 16.0.1 20260425 (prerelease) [gcc-16 r16-8812-gd9c07462a22]’
Flavor: [r-release-linux-x86_64](https://www.r-project.org/nosvn/R.check/r-release-linux-x86_64/redatam-00check.html)
```

These checks correspond to the 'rchk' subset. 

Because I maintain the 'redatam' package, I proposed a change in a new PR.

Ps. 'pugixml' has been an amazing tool to process Redatam files that have an XML component.

zeux:
I can reproduce this on GCC 16.0.1, looks like a regression from GCC 15. Full build error:

```
$ g++-16 --version
g++-16 (Ubuntu 16-20260322-1ubuntu1) 16.0.1 20260322 (experimental) [trunk r16-8246-g569ace1fa50]
[...]

PR #723, opened 2026-05-27
Proposed fix for GCC 16 warning

pachadotdev:
Solves https://github.com/zeux/pugixml/issues/721

### DIFF
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -1138,11 +1138,6 @@ namespace pugi
 	};
 }
 #else
-// CRAN patch: header is moved to the initializer list in xml_attribute_struct and
-// xml_node_struct below. GCC 16 -Wuninitialized flags PUGI_IMPL_GETPAGE(other)
-// as reading an uninitialized header when it is assigned in the constructor body.
-// PUGI_IMPL_GETHEADER_IMPL only does pointer arithmetic on `this`, so the move is safe.
-// Re-apply this patch after each pugixml vendor update.
 namespace pugi
 {
 	struct xml_attribute_struct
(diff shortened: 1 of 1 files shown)


zeux:
I can reproduce the original warning with g++-16 (using -O2 -Wall -Wextra) but not with g++-15. This patch can't be applied because it fails the builds, because using `this` in the initializer list produces warnings on other compilers.

It looks like the source of the warning is in `xml_document::_move`, at the very end where the new document is constructed. It also looks like this fixes it (and should be safe because `other_page` is equal to `PUGI_IMPL_GETPAGE(other)`):

```c++
diff --git a/src/pugixml.cpp b/src/pugixml.cpp
index 06bcb98..7ae9c19 100644
--- a/src/pugixml.cpp
+++ b/src/pugixml.cpp
@@ -7774,7 +7774,7 @@ namespace pugi
                }
 
                // reset other document
-               new (other) impl::xml_document_struct(PUGI_IMPL_GETPAGE(other));
+               new (other) impl::xml_document_struct(other_page);
                rhs._buffer = NULL;
        }
 #endif
```

zeux:
(feel free to update this PR with the suggested fix and remove the comments; I can merge this if the PR's diff is clean and it builds cleanly)

pachadotdev:
> (feel free to update this PR with the suggested fix and remove the comments; I can merge this if the PR's diff is clean and it builds cleanly)

thanks, I'll remove the comments

I added some extra changes after checking again vs R-devel with GCC 16

pachadotdev:
I edited it and it passes Windows tests now

zeux:
It looks like the new version simply silences the warning and does not apply a simpler change suggested in https://github.com/zeux/pugixml/pull/723#issuecomment-4556199447. Did that not work?

pachadotdev:
> It looks like the new version simply silences the warning and does not apply a simpler change suggested in [#723 (comment)](https://github.com/zeux/pugixml/pull/723#issuecomment-4556199447). Did that not work?

That was giving me some funky Windows warning. I see the change you made fixed that too. Thanks a lot!

PR #728, opened 2026-06-01
XPath: Extract string_value implementation for elements

zeux:
This is just extracting the existing code into a separate function and converting it to work with xml_node_struct*; this streamlines the codegen by removing a few redundant checks, and might make it easier to optimize in the future.

By itself this gives a ~5% speedup on string value heavy queries.


zeux:
AppVeyor failure is just due to a brief lapse in GH APIs returning merge status I believe, will ignore it for merging.

PR #731, opened 2026-06-16
Clarify operator precedence in declaration check

KirillStr:
Fixes a bug where mixing bitwise `&` and logical `&&` without explicit  parentheses caused incorrect operator precedence during XML header validation.


zeux:
What is the bug?

KirillStr:
Add explicit parentheses around the bitwise-AND group to clarify operator precedence and avoid compiler warnings when building with -Werror.

zeux:
What compiler are you using that gives the warning, with what options? Neither GCC nor Clang seem to warn on this with `-Wparentheses`. I'm trying to figure out if I should merge this before the release and rebuild and retest all release archives for the release tomorrow, or just merge it after; the report that omits this information is not ideal.

zeux:
Going to merge this a little later today as cleanup.

